PR #104, opened 2022-09-15
Modified the BLE example README files.

yc-adi:


lorne-maxim (approved):
Agree with Kevin on removing "Fitbit"

PR #131, opened 2022-10-03
Dual core sync example

yc-adi:
For MAX32655, projects RV_ARM_Loader and Hello_World are modified to demonstrate the synchronization between the ARM core and RISC-V core.


lorne-maxim (changes requested):
This is a great example.  I would ask that the DUAL_CORE_SYNC functionality be moved to its own example though instead of just adding it Hello_World.  Customers looking for a dual core example would probably never expect to find it as part of Hello_World.

lorne-maxim:
Sorry, just realized I said basically the same thing as Kevin.

If you've moved the code to a new project, please withdraw this PR.

PR #139, opened 2022-10-05
Rf test application

EdwinFairchild:
It was easier to just make a new PR with a new branch than to fix a bunch of conflict because of latest changes to main since I started this a while back. 

Still do know what is the range of TX dbm we can set... ? current FCC example has a max option of 4dbm and min of -10


kevin-gillespie (approved):
Looks good!

yc-adi:
Tested and works well.

yc-adi (approved):

victor-levi-maxim (approved):

PR #156, opened 2022-10-11
Explicitly Use clang-format-14

Jake-Carter:
Replace `Empty` values in .clang-format, which my result in variations between versions of the tool.


Jake-Carter:
@kevin-gillespie @EdwinFairchild 

The clang-format workflow should now match local runs with version 14.

Wall-E has been updated with v14 and a `CLANG_VERSION` variable has been added to the scripts.

See https://github.com/Analog-Devices-MSDK/msdk/pull/156/commits/ee88adcddc99d3e95f3c2f0831303927ae6245fc for latest auto-run

EdwinFairchild:
running clang-format --version on wall-e yields still v12. ? Or is it like another instance for github actions?

Jake-Carter:
> running clang-format --version on wall-e yields still v12. ? Or is it like another instance for github actions?

Use `clang-format-14 --version`.  

We can have the scripts call a specific version binary, but we can also symlink `clang-format` to `clang-format-14` for convenience on wall-e (which I have not done)

EdwinFairchild:
ah yeah i see the script now, idk why  i thought there was a space there and version was being passed as an argument.

lorne-maxim (approved):

EdwinFairchild (approved):

PR #173, opened 2022-10-17
Fix linkerfile search path

Jake-Carter:
Resolves https://jira.maxim-ic.com/browse/MSDK-903
- Fix linkerfile search path bug on MSYS2 when no MAXIM_PATH is defined
- Add LINKERPATH config variable
- LINKERPATH will fall back **to** the project directory if the linkerfile cannot be found

@ozersa This should fix the bug.  

MSYS2 will use its own paths (/c/maximsdk/...) if MAXIM_PATH is undefined, but it will still work.  I attempted to add a new [`fixpath`](https://github.com/Analog-Devices-MSDK/msdk/blob/0114df40f11c4bde30d945f33ca92148a6f4c8b3/Libraries/CMSIS/Device/Maxim/MAX32690/Source/GCC/gcc.mk#L258) in gcc.mk specifically for MSYS2 that used [`cygpath -m`](https://cygwin.com/cygwin-ug-net/cygpath.html).  However, it was too slow even if I did all the path calculation at the beginning of the file.  It added 2-3 seconds to the build, whereas this solution does not affect build time at all.


ozersa (dismissed):
Approved to merge

Jacob-Scheiffler (dismissed):

Examples/MAX32520/AES/Makefile
```diff
@@ -248,18 +251,20 @@ MFLOAT_ABI ?= softfp
...
 # Otherwise, gcc_riscv.mk sets the appropriate riscv linkerfile.
-LINKER = $(TARGET_LC).ld
-LINKERFILE ?= $(CMSIS_ROOT)/Device/Maxim/$(TARGET_UC)/Source/GCC/$(LINKER)
+LINKERFILE ?= $(TARGET_LC).ld
+LINKERPATH ?= $(CMSIS_ROOT)/Device/Maxim/$(TARGET_UC)/Source/GCC
+
+# Check if linkerfile exists
+ifeq ("$(wildcard $(LINKERPATH)/$(LINKERFILE))","")
+# Doesn't exists, attempt to use root project folder.
+LINKERPATH:=.
```
lorne-maxim, line 263: No need to hold up the PR merge for this, but assuming these files are auto-generated, you should change you script so this inserts "Doesn't exists..." instead of "Doesn't exist..." (remove the 's').

lorne-maxim (approved):

sihyung-maxim (approved):

PR #184, opened 2022-10-18
Added workaround for SRAM trim issue

sanjayjaroli:
ME10 has SRAM trim issue and requires setting gcr.scon[17:16]=2'b10 - for 1.1V +/- 10% on every power up which overwrites the wrong trims to the correct SRAM trims.

The ME14 also has the same issue and already has this implementation.


sihyung-maxim:
Please update your fork with the changes in main. The checks should all pass now once you do that.

lorne-maxim (approved):

PR #188, opened 2022-10-19
Hotfix: Use header_MAX32572.c

Jake-Carter:
This PR fixes a missed build error from https://github.com/Analog-Devices-MSDK/msdk/pull/180

The MAX32572 projects still need a significant amount of work and a migration to the new Makefile system, but this at least gets Hello_World building again.


sihyung-maxim (approved):
I'm hoping to get the ME55 PR, that's been hanging, merged in later today.

The CMSIS register files may still have some updates, so the migration to the new Makefile system can happen separately from the register updates.

lorne-maxim (approved):

PR #190, opened 2022-10-20
Examples: Fixed linter errors.

sihyung-maxim:
Fixed the linter errors in the newest examples that weren't checked in the first major pass-through.


sanjayjaroli (approved):
Good catch.

lorne-maxim (approved):

PR #201, opened 2022-10-24
ME21: Synced SVD files to register files.

sihyung-maxim:
Changes:
- Synced peripheral SVD files to match with register files.
- Updated copyright notices.
- Updated max32672.svd.
- Resolve SVDConv errors and warnings.
- Register files pass linter now.
- Deprecated unapproved SYS_AES references (not even used in UG). Officially, using AES name.
- Deprecated unapproved ECC register references (not even used in UG). Officially, using TRIMSIR name.
- Updated examples to remove deprecated names.


sihyung-maxim:
MXC_ECC is a subset of MXC_TRIMSIR registers. The "ECC" name should not have been added to the SDK when TRIMSIR was already present. Having multiple register files for a single peripheral (as in differently named peripherals sharing the same base location) causes an error with ARM's SVDConv utility since that keeps track of all the peripheral locations.

sihyung-maxim:
The SYS_AES name (instead of AES) was not approved for use in the MSDK and UG and was added without following the naming standard of the SVD files. I've deprecated that name and removed all reference to SYS_AES and replaced it with the AES name. The user guide will also be removing all references to SYS_AES if present in the current version as the peripheral/register/field names are shared between the MSDK and UG during autogeneration. Some may have slipped through the cracks.

sihyung-maxim:
The deprecated ECC and SYS_AES names are still supported in the MSDK, but all ME21-related files do not reference these deprecated names anymore. I've added deprecated warnings that will be noticed during build time, and I'll fully delete all references after a couple of releases.

Libraries/CMSIS/Device/Maxim/MAX32672/Include/sys_aes_regs.h
```diff
@@ -1,232 +0,0 @@
-/**
- * @file    sys_aes_regs.h
- * @brief   Registers, Bit Masks and Bit Positions for the SYS_AES Peripheral Module.
- */
-
-/* ****************************************************************************
```
Jacob-Scheiffler, line 6: Just double checking here, this isn't entirely deleted, right? It's just renamed to aes_regs.h?

sihyung-maxim: Yes, the new UG will have AES instead of SYS_AES. See comment above for more info.

Libraries/CMSIS/Device/Maxim/MAX32672/Include/sys_aes_key_regs.h
```diff
@@ -1,120 +0,0 @@
-/**
- * @file    sys_aes_key_regs.h
```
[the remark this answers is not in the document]
sihyung-maxim: Yes, this was changed to aes_key_regs.h. See comment above for more info.

Jacob-Scheiffler (approved):

lorne-maxim (approved):

PR #208, opened 2022-10-26
New Workflow: Check_Registers_SVD

sihyung-maxim:
Creating a new workflow to check whether both the register file and the corresponding peripheral SVD file are synced before merging if there are any register changes. This should keep our SVD files and User Guides up to date.


sihyung-maxim:
Need admin privileges to access private repos.

PR #216, opened 2022-10-29
[MSDK-934] Port ImgCapture and CameraIF_Debayer to AI87

Jake-Carter:
Depends on a MiscDrivers update:
* https://github.com/Analog-Devices-MSDK/Libraries-miscDrivers/pull/14

This PR ports the ImgCapture and CameraIF_Debayer examples to the AI87.  It also applies some misc. housekeeping to ImgCapture and CameraIF_Debayer for the AI85.


Examples/MAX78002/CameraIF_Debayer/utils.c
```diff
@@ -0,0 +1,177 @@
...
+
+#define DEBUG_COMPORT MXC_UART0
+
+/***************************** VARIABLES *************************************/
+
+/************************    PUBLIC FUNCTIONS  *******************************/
+//void utils_delay_ms(uint32_t ms)
+//{
+//    MXC_Delay(ms * 1000UL);
+//}
```
sihyung-maxim, line 54: Is this function still in development or should this be deleted?

sihyung-maxim (commented):
Looks good.

Examples/MAX78002/CameraIF_Debayer/utils.c
Jake-Carter: Thanks, this is one leftover from the original CameraIF example that can be deleted.  Removing it now

Examples/MAX78002/CameraIF_Debayer/utils.c
```diff
@@ -0,0 +1,173 @@
...
+
+/************************    PUBLIC FUNCTIONS  *******************************/
+
+uint32_t utils_get_time_ms(void)
+{
+    int sec;
+    double subsec;
+    uint32_t ms;
+
+    subsec = MXC_RTC_GetSubSecond() / 4096.0;
```
lorne-maxim, line 58: Consider doing this with integer math so the floating point libraries are not linked in (unless used somewhere else in the application).

uint32_t subsec = (MXC_RTC_GetSubSecond() * 1000) / 4096;

Also be aware subsecond count could rollover between calls to GetSubSecond and GetSecond.  Back to back calls to this function could show time going backwards :)  Isn't there a GetTime function in the RTC API that performs an "atomic" read of the two registers?

Examples/MAX78002/CameraIF_Debayer/.gitignore
```diff
@@ -0,0 +1 @@
+pc_utility/*.png
```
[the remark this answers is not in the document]
Jake-Carter: Yessir, the utility will save some output image files that I have accidentally committed in the past

Examples/MAX78002/CameraIF_Debayer/utils.c
Jake-Carter: Thanks, I inherited these functions from the old CameraIF example and didn't notice this one.  The code doesn't use it at all, so I just removed it.

lorne-maxim (approved):

sihyung-maxim (approved):

PR #218, opened 2022-11-01
Dev examples return value update

karaanil:
MSDK one shot examples return value update pull request.

### DIFF
--- a/Examples/MAX32520/AES/main.c
+++ b/Examples/MAX32520/AES/main.c
@@ -416,11 +416,10 @@ int main(void)
 
     if (fail == 0) {
         printf("\nExample Succeeded\n");
-        return 0;
     } else {
         printf("\nExample Failed\n");
-        return -1;
+        return E_FAIL;
     }
 
-    return 0;
+    return E_NO_ERROR;
 }
--- a/Examples/MAX32520/AES/main.c
+++ b/Examples/MAX32520/AES/main.c
@@ -416,8 +416,10 @@ int main(void)
 
     if (fail == 0) {
         printf("\nExample Succeeded\n");
+        return 0;
     } else {
         printf("\nExample Failed\n");
+        return -1;
     }
 
     return 0;
--- a/Examples/MAX32520/DMA/main.c
+++ b/Examples/MAX32520/DMA/main.c
@@ -219,11 +219,10 @@ int main(void)
 
     if (fail == 0) {
         printf("\nExample Succeeded\n");
-        return 0;
     } else {
         printf("\nExample Failed\n");
-        return -1;
+        return E_FAIL;
     }
 
-    return 0;
+    return E_NO_ERROR;
 }
--- a/Examples/MAX32520/DMA/main.c
+++ b/Examples/MAX32520/DMA/main.c
@@ -97,20 +97,19 @@ void example1(void)
     if (retval != E_NO_ERROR) {
         printf("Failed MXC_DMA_Init().\n");
         fail += 1;
-    }
-    else {
-    	flag = 0;
-		MXC_DMA_MemCpy(dstdata, srcdata, MAX_SIZE, memCpyComplete);
+    } else {
+        flag = 0;
+        MXC_DMA_MemCpy(dstdata, srcdata, MAX_SIZE, memCpyComplete);
 
-		while (flag == 0) {}
+        while (flag == 0) {}
 
-		//Validate
-		if (memcmp(srcdata, dstdata, MAX_SIZE) != 0) {
-			printf("Data mismatch.\n");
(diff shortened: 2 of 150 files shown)


Examples/MAX32650/SPI/main.c
```diff
@@ -221,5 +221,5 @@ int main(void)
         LED_On(1);
     }
 
-    return E_NO_ERROR;
+    return 0;
```
Jake-Carter, line 224: We should keep `return E_NO_ERROR` instead of using `return 0` for all main functions

Examples/MAX32650/UART/main.c
```diff
@@ -128,16 +128,14 @@ int main(void)
     if (error < E_NO_ERROR) {
         printf("-->Error initializing UART: %d\n", error);
         printf("-->Example Failed\n");
-
-        while (1) {}
+        return -1;
```
Jake-Carter, line 131: Same for here.  Instead of using `return -1` we should use `return error` (in this case) or add a generic `E_FAIL` to `mxc_errors.h`.  We should avoid returning "naked" hard-coded files like 0 or -1.

Examples/MAX32655/I2C/main.c
```diff
@@ -64,7 +64,7 @@
 #define I2C_SLAVE_ADDR (0x51)
 #define I2C_BYTES 100
 
-typedef enum { FAILED, PASSED } test_t;
+typedef enum { FAILED = -1, PASSED = 0 } test_t;
```
Jake-Carter, line 67: My suggestions are similar to what is done for this example here, except that we should move the functionality to `mxc_errors.h`.

Jake-Carter (approved):
Looks good!  Please merge main into your branch to resolve merge conflicts.

For a future PR:

We should add a generic `#define E_FAIL -255` to `mxc_errors.h`.  

Then, instead of using `return 0` we should use `return E_NO_ERROR`, and instead `return -1` we should use `return E_FAIL`.  This improves readability and maintainability.  It easily differentiates error codes in return values.  It also will allow the API testing setup much more flexibility in the future.

However, this PR is a massive improvement in the right direction.  So there's no need to address it here unless you feel like doing it now.

Examples/MAX32520/AES/main.c
```diff
@@ -416,8 +416,10 @@ int main(void)
 
     if (fail == 0) {
         printf("\nExample Succeeded\n");
+        return 0;
     } else {
         printf("\nExample Failed\n");
+        return -1;
     }
 
     return 0;
```
sihyung-maxim, line 425: This can be removed, it'll never reach here.

sihyung-maxim (commented):
There is a return 0 line at the end of main.c that will never be reached in many of these examples. Is that something we want to delete?

The MAX32570 (ME13) file changes should also be done in the Examples-MAX32570 submodule.

Jake-Carter:
@ozersa @karaanil Friendly reminder!  Is this still a WIP?  Have you had a chance to review our comments above?

Examples/MAX32520/DMA/main.c
```diff
@@ -96,24 +96,21 @@ void example1(void)
...
-
-        while (1) {}
-
         fail += 1;
-    } else {
-        printf("Data verified.\n");
+    }
+    else {
+    	flag = 0;
+		MXC_DMA_MemCpy(dstdata, srcdata, MAX_SIZE, memCpyComplete);
```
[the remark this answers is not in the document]
karaanil: @Jake-Carter @ozersa , I will change all return values from 0 and -1 to E_NO_ERROR and E_FAIL. I will also run clang formatter for each example modified in this branch. It might be better to merge them after such change.

Examples/MAX32520/AES/main.c
karaanil: @sihyung-maxim @ozersa I will remove such unnecessary return values on the examples modified in this branch. I will merge it after such change.

karaanil:
@Jake-Carter 

- I have run all actions required and currently this branch passes all of them.
- I have rebased it onto the up-to-date main branch.

However, I have closed this pull request to make it not to send e-mail while I am applying changes and currently I can not reopen it.
How can we reopen this pull request?
The button at the bottom is disabled.

Jake-Carter:
@karaanil I renamed it and the button appeared again, but it won't let me reopen it.  I'm not sure why, but the force-push prevents it.  I recreated it in https://github.com/Analog-Devices-MSDK/msdk/pull/317

karaanil:
@Jake-Carter thanks, I have approved #317.
Note: I needed to force push for rebase on the up-to-date main.

PR #223, opened 2022-11-04
Update `bin2wav.py' script

maximreza:
- Add `-a` option to convert all sample audio files to wave in all subdirectories while skipping the ones that have already been converted.
- Update README.md file accordingly.


Jake-Carter (approved):
Looks good

PR #229, opened 2022-11-09
Adding ST7735 TFT Controller driver.

Jacob-Scheiffler:
Adding ST7735 driver collateral in Boards/MAX32690/EvKit_V1.


Libraries/Boards/MAX32690/EvKit_V1/Source/board.c
```diff
@@ -131,6 +149,66 @@ int Console_PrepForSleep(void)
...
+            } else {
+                if (tx_byte & 0x80) { //Shift out data MSB first
+                    tft_mosi.port->out_set = tft_mosi.mask;
+                } else {
+                    tft_mosi.port->out_clr = tft_mosi.mask;
+                }
+                tx_byte = tx_byte << 1;
+            }
+
+            for (int k = 0; k < 2; k++) {}
```
lorne-maxim, line 198: These delay loops may not work if the compiler decides to be "smart" and optimize them away.  Does MXC_Delay not have good enough resolution?  A set of "__NOP();" instructions may be better, but then you're subject to whatever the system clock is running at.

lorne-maxim: Had a hallway conversation with Jacob about this.  MXC_Delay doesn't offer a short enough delay due to the overhead it has of converting the passed in time to system tick values.  Removing these delays altogether also works, but results in a non-50% duty cycle.  Since this file is board specific anyway, we decided to go ahead and leave these SW delay loops as-is.

lorne-maxim (changes requested):
Please also update Libraries/Boards/MAX32690/EvKit_V1/examples.txt with the new example.

Jake-Carter (approved):
Looks good!  Just throw a comment in the implementation for the SPI functions about them being bit-bang

lorne-maxim (approved):

PR #233, opened 2022-11-14
MSDK-952: Fix TFT TS issue

ozersa:
TFT Demo TS issue fixed, 
Tested on MAX78000 EvKit RevC.

If there is some other TFT example which TS does not work it most probably can be fix by this type change.

Signed-off-by: Sadik.Ozer <sadik.ozer@analog.com>

### DIFF
--- a/Examples/MAX78000/TFT_Demo/example_config.h
+++ b/Examples/MAX78000/TFT_Demo/example_config.h
@@ -34,6 +34,8 @@
 #ifndef EXAMPLES_MAX78000_TFT_DEMO_EXAMPLE_CONFIG_H_
 #define EXAMPLES_MAX78000_TFT_DEMO_EXAMPLE_CONFIG_H_
 
+#include "board.h"
+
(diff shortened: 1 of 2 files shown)


lorne-maxim (approved):

Examples/MAX78000/TFT_Demo/example_config.h
```diff
@@ -34,6 +34,8 @@
 #ifndef EXAMPLES_MAX78000_TFT_DEMO_EXAMPLE_CONFIG_H_
 #define EXAMPLES_MAX78000_TFT_DEMO_EXAMPLE_CONFIG_H_
 
+#include "board.h"
```
Jake-Carter, line 37: Wow, good catch!  "board.h" was never included, so the ifdefs below were never triggered...  missing `ENABLE_TS` entirely.

Jake-Carter (approved):
Thanks @ozersa

PR #234, opened 2022-11-14
Converted all submodules except Cordio to be internal to the repo.

lorne-maxim:


Jake-Carter (approved):
Looks good!

A `git submodule deinit --force --all` is required to pull this branch, so we should notify developers of this change before merging

sihyung-maxim (approved):

PR #238, opened 2022-11-16
Add Touch Screen Support for AI87 EVKIT + NewHaven TFT

Jake-Carter:
This PR adds touchscreen support for the AI87 EVKIT (tsc2046.c)

The SPI transport layer had to be moved out of the tsc2046.c file and into the board files.  The AI87 EVKIT requires software control of the Chip Select pin, and additionally does not have the TS_BUSY signal routed, so its SPI transport function differs completely from the old one.

Currently, the drivers and hardware works but there is a mismatch between the TFT display's coordinate system and the touchscreen's.  This can most likely be fixed in software by post-processing the X,Y coordinates returned from the touchscreen to match the TFT.  I've already tried swapping x,y values alongside various `FLIP_SCREEN` and `ROTATE_SCREEN` combinations, but haven't successfully matched the correct coordinates yet.

### DIFF
--- a/Examples/MAX78002/TFT_Demo/main.c
+++ b/Examples/MAX78002/TFT_Demo/main.c
@@ -38,18 +38,12 @@
 #include <string.h>
 #include "mxc.h"
 
-#define TOD_START_TIME (12 * SECS_PER_HR + 34 * SECS_PER_MIN + 56)
-#define TOD_ALARM_INTERVAL 2
-#define SECS_PER_MIN 60
-#define SECS_PER_HR (60 * SECS_PER_MIN)
-#define SECS_PER_DAY (24 * SECS_PER_HR)
 #define TS_X_MIN 254
 #define TS_X_MAX 3680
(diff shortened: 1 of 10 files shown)


lorne-maxim (changes requested):
Looks like the SPI functions in board.c need to be excluded when building for the RiscV.  SPI0 is on the AHB and not available to that core.

Jake-Carter:
WIP:  MAX32655 also uses tsc2046, need to add transmit functions for its board files too

lorne-maxim (approved):

Jake-Carter:
Tested and functional on:
- [x] AI85 EVKIT (TFT_Demo)
- [x] AI87 EVKIT (TFT_Demo)
- [x] ME17 EVKIT (TFT_Demo)

**WIP**: Match TouchScreen coordinate system to TFT coordinate system on AI87

sihyung-maxim (approved):

PR #240, opened 2022-11-16
Add I2S_DMA_Target Example for AI85

Jake-Carter:
This PR adds max9867 codec drivers to a new MiscDrivers folder (CODEC) and updates the AI85 FTHR board makefile to add them to the build (There is no codec on the EVKIT).

It also adds a new example for simultaneous recording and playback from the codec using DMA from @jessexm


Examples/MAX78000/I2S_DMA_Target/main.c
```diff
@@ -0,0 +1,224 @@
...
+        /*
+      non-dma activity triggered work
+    */
+    }
+}
+
+void i2c_init(void)
+{
+    if (MXC_I2C_Init(CODEC_I2C, 1, 0))
+        blink_halt("Error initializing I2C controller");
```
lorne-maxim, line 145: Explicitly comparing the return value against E_NO_ERROR would be better.

Libraries/MiscDrivers/CODEC/max9867.c
```diff
@@ -0,0 +1,431 @@
...
+
+  return MXC_I2C_MasterTransaction(&i2c_req);
+}
+
+static int reg_update(uint8_t reg, uint8_t mask, uint8_t val)
+{
+  int err;
+  uint8_t tmp;
+
+  if ((err = reg_read(reg, &tmp)))
```
lorne-maxim, line 102: Compare result to E_NO_ERROR

Libraries/MiscDrivers/CODEC/max9867.c
```diff
@@ -0,0 +1,431 @@
...
+{
+  return reg_update(MAX9867_16_MODE, 0x07, mode);
+}
+
+int max9867_power_enable(int shutdown, int enable)
+{
+  int err;
+
+  if (shutdown)
+    if ((err = reg_write(MAX9867_17_PWR_SYS, 0x00)))
```
lorne-maxim, line 308: Is the intent if shutdown succeeds to move on to the 'enable' settings.  Seems like maybe the shutdown block should always return?

lorne-maxim (changes requested):
There are lots of if statements like:

```
if(err = func()) {
  return err;
}
```

I think these are better if they explicitly compare against E_NO_ERROR:

```
if((err = func()) != E_NO_ERROR) {
  return err;
}
```

lorne-maxim:
Also, don't forget to add the new example to examples.txt for all boards where the example is expected to work.

Examples/MAX78000/I2S_DMA_Target/main.c
Jake-Carter: Fixed in https://github.com/Analog-Devices-MSDK/msdk/pull/240/commits/1c5ae59594fba5650b3ca48105437a28f0cc9f0f

Libraries/MiscDrivers/CODEC/max9867.c
Jake-Carter: Addressed in https://github.com/Analog-Devices-MSDK/msdk/pull/240/commits/1c5ae59594fba5650b3ca48105437a28f0cc9f0f

sihyung-maxim (approved):

Jacob-Scheiffler (approved):

lorne-maxim (approved):
There are still some things that I would have though clang-format would have changed.  Most significantly is the single line if blocks without braces.  No need to block this PR, but we should make sure all the old sub-module files are clang-formatted properly before the next release.

PR #243, opened 2022-11-17
Configure ERTCO enable bit.

lorne-maxim:
The B revisions of the MAX32670 and MAX32672 have a new RTC enable bit in the PWRSEQ register.  This PR properly configures this bit when enabling/disabling the ERTCO.

### DIFF
--- a/Libraries/PeriphDrivers/Source/SYS/sys_me21.c
+++ b/Libraries/PeriphDrivers/Source/SYS/sys_me21.c
@@ -151,6 +151,7 @@ void MXC_SYS_ClockEnable(mxc_sys_periph_clock_t clock)
 /* ************************************************************************** */
 void MXC_SYS_RTCClockEnable()
 {
+    MXC_PWRSEQ->lpcn &= ~(MXC_F_PWRSEQ_LPCN_TM_PWRSEQ);
     MXC_GCR->clkctrl |= MXC_F_MCR_CLKCTRL_ERTCO_EN;
 }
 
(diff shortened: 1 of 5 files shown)


sihyung-maxim (approved):

lorne-maxim:
I'm not sure why the build is failing.  It looks like the necessary #include statements are in place.  Investigating now...

sihyung-maxim:
> I'm not sure why the build is failing. It looks like the necessary #include statements are in place. Investigating now...

Yes, the PWRSEQ_LPCN ERTCO_PD and ERTCO_EN fields are present in the ME15's pwrseq_regs.h...

Libraries/PeriphDrivers/Source/SYS/sys_me21.c
```diff
@@ -160,6 +163,8 @@ int MXC_SYS_RTCClockDisable(void)
     /* Check that the RTC is not the system clock source */
     if ((MXC_GCR->clkctrl & MXC_F_GCR_CLKCTRL_SYSCLK_SEL) != MXC_S_GCR_CLKCTRL_SYSCLK_SEL_ERTCO) {
         MXC_GCR->clkctrl &= ~MXC_F_MCR_CLKCTRL_ERTCO_EN;
```
Jacob-Scheiffler, line 165: This may be an error in the UG and not in the code, but should this be clearing a bit in MCR? Or should there be an added field in GCR->clkctrl?

lorne-maxim: Good catch!  Just pushed a change that writes to the MXC_MCR->clkctrl instead of MSC_GCR->clkctrl.  Luckily bit 17 is undefined in GCR, so the old code wasn't doing anything bad.

Jacob-Scheiffler (approved):

PR #254, opened 2022-11-19
Set -O2 as Default for AI85, AI87 

Jake-Carter:
This PR sets the default optimization level to `-O2` for the AI85 and AI87.

Projects interfacing with camera drivers also have their optimization level explicitly set to `-O2` in project.mk to increase visibility, since it's a hard requirement for the DMA timing.

It also replaces the example for setting `BOARD` with an example for setting the optimization level to `-Og`, since this was causing some confusion.  VS Code and Eclipse set `BOARD` on the command-line and have their own settings for changing it, which takes higher [precedence](https://github.com/Analog-Devices-MSDK/VSCode-Maxim/blob/develop/README.md#precedence-hierarchy).  Future support for explicit debug builds will be improved in these IDEs in order to address this confusion more directly.

### DIFF
--- a/Examples/MAX78000/ADC/Makefile
+++ b/Examples/MAX78000/ADC/Makefile
@@ -233,9 +233,7 @@ ifeq ($(DEBUG),1)
 MXC_OPTIMIZE_CFLAGS := -Og
 endif
 
-# Fallback default optimizes for debugging as recommended
-# by GNU for code-edit-debug cycles
-# https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#Optimize-Options
+# Set default optimization level -O2 to maximize performance for CNN code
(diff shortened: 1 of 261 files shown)


lorne-maxim (approved):

Saicharan036:
Hi Team, 

I am using this SPI example for reading the Chip id of the device, There is some extra pulse appearing in at the beginning of the clock pulse as shown in the below image, please tell me what could be the reason.

image

maximreza:
this could be an artifact. it is possible that both pins are charging up slowly and hit the Saleae threshold. Could you try the analog mode on the channels and see what is the actual shape of the signal?

Jake-Carter:
@Saicharan036 @maximreza I have transferred this to #305

PR #267, opened 2022-11-28
Ignoring OBW tests on higher channels.

kevin-gillespie:


tim-magnusen (approved):
This will probably have to be revisited with WLP since the frequency pulling is larger.

kevin-gillespie:
I'm working on installing a WLP board to our test setup. 

Another thing I've noticed is that the packet length affects the OBW. With a 0 length payload, we transmit less at the channel frequency, causing the OBW to increase.

PR #276, opened 2022-12-02
Porting the SPIXF_ICC example to MAX32665

Jacob-Scheiffler:
This PR adds the SPIXF_ICC example, originally written for the MAX32650, to the MAX32665. To make it work, I needed to add the same ICC support that was added for the MAX32650 which allows for a single cache controller instance to be operated on at a time. That allowed the SFCC to be enabled and disabled without affecting the state of either ICC0 or ICC1.


Examples/MAX32665/SPIXF_SFCC/main.c
```diff
@@ -0,0 +1,253 @@
...
+        printf("Examples failed.\n");
+        return E_BAD_STATE;
+    }
+    MXC_RTC_Start();
+
+    // Initialize interface (SPIXF) with external flash chip
+    if (Ext_Flash_Init() != E_NO_ERROR) {
+        printf("External flash initialization Failed\n");
+        printf("Example Failed\n");
+        while (1) {}
```
Jake-Carter, line 179: Return error code instead of blocking

Examples/MAX32665/SPIXF_SFCC/main.c
```diff
@@ -0,0 +1,253 @@
...
+    }
+    printf("%s Initialized.\n", EXT_FLASH_NAME);
+    Ext_Flash_Reset();
+
+    // Check ID of the external flash
+    if ((id = Ext_Flash_ID()) != EXT_FLASH_EXP_ID) {
+        printf("Error verifying external flash ID: 0x%x\n", id);
+        printf("Example Failed\n");
+        while (1) {}
```
Jake-Carter, line 188: Return error code instead of blocking

Examples/MAX32665/SPIXF_SFCC/main.c
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    // Erase sector of external flash to store test function
+    if ((err = Ext_Flash_Erase(0x00000, Ext_Flash_Erase_64K)) != E_NO_ERROR) {
+        printf("Flash erase failed with error code: %d\n", err);
+        printf("Example Failed\n");
+        while (1) {}
```
Jake-Carter, line 195: Return error code instead of blocking

Examples/MAX32665/SPIXF_SFCC/main.c
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    // Enable Quad mode if necessary
+    if (EXT_FLASH_SPIXFC_WIDTH == Ext_Flash_DataLine_Quad) {
+        if (Ext_Flash_Quad(1) != E_NO_ERROR) {
+            printf("Error enabling quad mode\n\n");
+            while (1) {}
```
Jake-Carter, line 202: Return error code instead of blocking

Examples/MAX32665/SPIXF_SFCC/main.c
```diff
@@ -0,0 +1,253 @@
...
+        }
+    } else {
+        if (Ext_Flash_Quad(0) != E_NO_ERROR) {
+            printf("Error disabling quad mode\n\n");
+            while (1) {}
```
Jake-Carter, line 207: Return error code instead of blocking

Examples/MAX32665/SPIXF_SFCC/main.c
```diff
@@ -0,0 +1,253 @@
...
+    // Load test function into external flash and set function pointer to it's location
+    printf("Loading test function into external flash.\n\n", (uint32_t)(&__load_length_xip),
+           &__load_start_xip);
+    if ((err = Ext_Flash_Program_Page(EXT_FLASH_ADDR, &__load_start_xip,
+                                      (uint32_t)(&__load_length_xip), EXT_FLASH_SPIXFC_WIDTH)) !=
+        E_NO_ERROR) {
+        printf("Error Programming: %d\n", err);
+        fail++;
+    }
+    func = (void (*)(void))(MXC_XIP_MEM_BASE | 0x1);
```
Jake-Carter, line 220: It'd be worth adding some comments explaining how this function pointer assignment works

Examples/MAX32665/SPIXF_SFCC/main.c
```diff
@@ -0,0 +1,253 @@
...
+
+    // Setup SPIX
+    spixf_cfg_setup();
+
+    printf("Setup complete. Press SW2 to run SFCC test.\n\n");
```
Jake-Carter, line 225: Is SW2 correct for all possible boards?

Libraries/PeriphDrivers/Include/MAX32665/icc.h
```diff
@@ -85,6 +85,32 @@ void MXC_ICC_Disable(void);
  */
 void MXC_ICC_Flush(void);
 
+/**
+ * @brief   Reads ID information from one of the ICC's Cache ID Register.
+ * @param   icc     Pointer ICC instance to get ID information from.
+ * @param   cid     Selects what information to get from the Cache ID Register
+ * @retval  Returns the selected value from the Cache ID Register.
+ */
+int MXC_ICC_IDInst(mxc_icc_regs_t *icc, mxc_icc_info_t cid);
```
Jake-Carter, line 94: Consider...
```C
int MXC_ICC_GetID(mxc_icc_regs_t *icc, mxc_icc_info_t cid);
````
instead.  The other function name `MXC_ICC_ID` is terrible so in this case we should make this one the standard.

**Review by Jake-Carter — CHANGES_REQUESTED** [review]
Looks good!  A few minor things below

File: Examples/MAX32665/SPIXF_SFCC/main.c (hunk 7 above)
Comment (Jacob-Scheiffler) [reply Examples/MAX32665/SPIXF_SFCC/main.c:224]: Yes and no. The push button at index 0 is not labeled "SW2" on every EV kit, but this example is only compatible with the MAX32665 at the moment. So it is correct for all possible boards right now, but there could be another chip in the future that is compatible with this example and it does not call PB0 "SW2".

Comment (Jacob-Scheiffler) [reply Examples/MAX32665/SPIXF_SFCC/main.c:224]: Other examples also have a pushbutton name hardcoded in the printf, maybe we should define those in board.h?

File: Libraries/PeriphDrivers/Include/MAX32665/icc.h (hunk 8 above)
Comment (Jacob-Scheiffler) [reply Libraries/PeriphDrivers/Include/MAX32665/icc.h:94]: I agree that that's a much better name, but I want it to be consistent with what's in MAX32650/icc.h. We could deprecate it? It's only been around for a month or two.

File: Examples/MAX32665/SPIXF_SFCC/project.mk [hunk 9]
```diff
@@ -0,0 +1,17 @@
...
+# For instructions on how to use this system, see
+# https://github.com/Analog-Devices-MSDK/VSCode-Maxim/tree/develop#build-configuration
+
+#BOARD=FTHR_RevA
+# ^ For example, you can uncomment this line to make the 
+# project build for the "FTHR_RevA" board.
+
+# **********************************************************
+
+# Add your config here!
```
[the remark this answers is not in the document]
Jake-Carter: Cool, no change needed then.  Just wanted to check - it's bitten me in the past.

Examples/MAX32665/SPIXF_SFCC/main.c
Jake-Carter: Sounds good.  No change needed then.  Agreed - we should work out a long-term solution in board.h to associate a string with the pushbuttons.  It's a common enough gotcha to make it worth it.

Libraries/PeriphDrivers/Include/MAX32665/icc.h
Jake-Carter: Ah - thought you were creating a new function.  If `MXC_ICC_IDInst` already exists then we're forced to use it.  No change needed.

Examples/MAX32665/SPIXF_SFCC/project.mk
```diff
@@ -0,0 +1,17 @@
+# This file can be used to set build configuration
+# variables.  These variables are defined in a file called 
+# "Makefile" that is located next to this one.
+
+# For instructions on how to use this system, see
+# https://github.com/Analog-Devices-MSDK/VSCode-Maxim/tree/develop#build-configuration
+
+#BOARD=FTHR_RevA
```
Jake-Carter, line 8: Just noticed this - Should be `BOARD=FTHR` or `BOARD=FTHR2` for MAX32665.

You can just delete this whole comment header and replace it with the latest one:

https://github.com/Analog-Devices-MSDK/VSCode-Maxim/blob/65af7c61800c7039956f3c1971ffd7915008668d/MaximSDK/Inject/project.mk

I removed the BOARD example because it was causing too much confusion.

Jake-Carter (changes requested):
Looks good!  Thanks.  Caught one last thing in project.mk.  Fix that and it's ready for merge.

Jake-Carter (approved):

lorne-maxim (approved):

PR #279, opened 2022-12-05
Workflows: clang-format-run: Fixed directory exclusions in find command.

sihyung-maxim:
Fixed regex pattern in find command for clang-format-run.sh.


Jacob-Scheiffler (approved):

kevin-gillespie (approved):

Jake-Carter (changes requested):
We should make the runner match the checker exactly:

https://github.com/Analog-Devices-MSDK/msdk/blob/f1942550e0d9ce5cc0cb3727cdfe15879d3bef40/.github/workflows/clang-format-check.yml#L56

CNN generated files should be skipped

Jake-Carter (approved):
Looks good!  Thanks @sihyung-maxim

PR #280, opened 2022-12-05
PTG: New interrupt registers for ME12, ME18, and ME55.

sihyung-maxim:
There was an updated version of the PT registers when the ME18 was introduced. There are two new registers called: READY_INTFL (0x018) and READY_INTEN (0x01C). The original INTFL (0x008) and INTEN (0x00C) registers were renamed to STOP_INTFL (0x008) and STOP_INTEN (0x00C). The user guides currently don't have these new registers, but Clayton will add them into the next UG version.

INTFL -> STOP_INTFL (0x0008)
INTEN -> STOP_INTEN (0x000C)
READY_INTFL (0x0018)
READY_INTEN (0x001C)

What do you all think about these new names? The STOP\_ and READY\_ prefixes felt more descriptive even though registers should generally avoid having "_" within their names. There's also the issue with code compatibility with changing the original INTFL/INTEN names, but I think adding the prefixes would be helpful to the users. I'm open to renaming the interrupt registers to something else or leaving the original INTFL/INTEN register names as is if it causes too much conflict.


Libraries/PeriphDrivers/Source/PT/pt_reva.c
```diff
@@ -148,22 +148,62 @@ void MXC_PT_RevA_SetPattern(unsigned pts, uint32_t pattern)
...
+    return ptg->stop_intfl;
+}
+
+void MXC_PT_RevA_ClearStopFlags(mxc_ptg_reva_regs_t *ptg, uint32_t flags)
+{
+    ptg->stop_intfl = flags;
+}
+
+void MXC_PT_RevA_EnableReadyInt(mxc_ptg_reva_regs_t *ptg, uint32_t pts)
+{
```
Jacob-Scheiffler, line 190: These functions should use ptg->ready_intfl

Jacob-Scheiffler (changes requested):

Libraries/PeriphDrivers/Include/MAX32662/pt.h
```diff
@@ -179,7 +179,7 @@ void MXC_PT_SetPattern(unsigned pts, uint32_t pattern);
 void MXC_PT_EnableInt(uint32_t pts);
 
 /**
- * @brief      Disable interrupts for the pulse trains selected.
+ * @brief      Disable Stop interrupts for the pulse trains selected.
```
Jacob-Scheiffler, line 182: This goes for other chips and functions as well, could you add a comment explaining that this function is included for compatibility and that users should use MXC_PT_DisableStopInt() instead?

Libraries/PeriphDrivers/Source/PT/pt_reva.c
```diff
@@ -43,7 +43,7 @@ void MXC_PT_RevA_Init(mxc_ptg_reva_regs_t *ptg, mxc_clk_scale_t clk_scale)
     ptg->enable = 0;
 
     //clear all interrupts
-    ptg->intfl = ptg->intfl;
```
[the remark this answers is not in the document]
sihyung-maxim: Thanks. Just a note, even though parts without the ready_intfl/ready_inten registers will call the Init function, functionally, nothing should happen.

Libraries/PeriphDrivers/Include/MAX32662/pt.h
sihyung-maxim: Should we add the deprecated attribute for the EnableInt, DisableInt, ClearFlags, and GetFlags functions in favor of the new Stop and Ready functions? Or just leave it as is and add the note? (Assuming we're going forward with using STOP_INTFL/STOP_INTEN instead of INTFL/INTEN)

Jacob-Scheiffler: I would think it's best practice to add deprecated? Does adding just cause a warning during compilation if someone tries to use that function?

Libraries/CMSIS/Device/Maxim/MAX32572/Include/ptg_regs.h
```diff
@@ -179,67 +179,67 @@ typedef struct {
...
  * @ingroup  ptg_registers
- * @defgroup PTG_INTFL PTG_INTFL
- * @brief    Pulse Train Interrupt Flags
+ * @defgroup PTG_STOP_INTFL PTG_STOP_INTFL
+ * @brief    Pulse Train Stop Interrupt Flags
  * @{
  */
-#define MXC_F_PTG_INTFL_PT0_POS                        0 /**< INTFL_PT0 Position */
-#define MXC_F_PTG_INTFL_PT0                            ((uint32_t)(0x1UL << MXC_F_PTG_INTFL_PT0_POS)) /**< INTFL_PT0 Mask */
+#define MXC_F_PTG_STOP_INTFL_PT0_POS                   0 /**< STOP_INTFL_PT0 Position */
```
Jake-Carter, line 186: These offsets are more of a pain in the ass to redirect.  I would argue we should do it for backward compatibility, but only if the register file generator can do it.  Is it feasible to add a feature like that to the generator?

Libraries/PeriphDrivers/Include/MAX32662/pt.h
Jake-Carter: Agreed, add deprecated.  Yes, it counts as a warning.  However, the warning is thrown whenever somebody compiles the file, _not_ when someone uses the function.  So it's not trivial.  We will need some sort of compiler definitions to wrap the deprecated functions in to avoid the warnings every time the PT drivers are compiled.

See [nvic_table.h](https://github.com/Analog-Devices-MSDK/msdk/blob/main/Libraries/PeriphDrivers/Include/MAX32520/nvic_table.h) for an example

Libraries/CMSIS/Device/Maxim/MAX32662/Include/ptg_regs.h
```diff
@@ -103,10 +105,12 @@ typedef struct {
  */
 #define MXC_R_PTG_ENABLE                   ((uint32_t)0x00000000UL) /**< Offset from PTG Base Address: <tt> 0x0000</tt> */
 #define MXC_R_PTG_RESYNC                   ((uint32_t)0x00000004UL) /**< Offset from PTG Base Address: <tt> 0x0004</tt> */
-#define MXC_R_PTG_INTFL                    ((uint32_t)0x00000008UL) /**< Offset from PTG Base Address: <tt> 0x0008</tt> */
```
Jake-Carter, line 106: Redirect, not delete

Libraries/CMSIS/Device/Maxim/MAX32662/Include/ptg_regs.h
```diff
@@ -103,10 +105,12 @@ typedef struct {
...
-#define MXC_R_PTG_INTEN                    ((uint32_t)0x0000000CUL) /**< Offset from PTG Base Address: <tt> 0x000C</tt> */
```
Jake-Carter, line 107: Redirect, not delete

Libraries/CMSIS/Device/Maxim/MAX32662/Include/ptg_regs.h
```diff
@@ -151,43 +155,43 @@ typedef struct {
 
 /**
  * @ingroup  ptg_registers
- * @defgroup PTG_INTFL PTG_INTFL
- * @brief    Pulse Train Interrupt Flags
+ * @defgroup PTG_STOP_INTFL PTG_STOP_INTFL
+ * @brief    Pulse Train Stop Interrupt Flags
  * @{
  */
-#define MXC_F_PTG_INTFL_PT0_POS                        0 /**< INTFL_PT0 Position */
```
Jake-Carter, line 158: Same comments as other reg files - redirect these but only if the generator can do it.

Libraries/CMSIS/Device/Maxim/MAX32690/Include/ptg_regs.h
```diff
@@ -103,10 +105,12 @@ typedef struct {
  */
 #define MXC_R_PTG_ENABLE                   ((uint32_t)0x00000000UL) /**< Offset from PTG Base Address: <tt> 0x0000</tt> */
 #define MXC_R_PTG_RESYNC                   ((uint32_t)0x00000004UL) /**< Offset from PTG Base Address: <tt> 0x0004</tt> */
-#define MXC_R_PTG_INTFL                    ((uint32_t)0x00000008UL) /**< Offset from PTG Base Address: <tt> 0x0008</tt> */
```
Jake-Carter, line 106: Redirect, not delete

Libraries/CMSIS/Device/Maxim/MAX32690/Include/ptg_regs.h
```diff
@@ -103,10 +105,12 @@ typedef struct {
...
-#define MXC_R_PTG_INTEN                    ((uint32_t)0x0000000CUL) /**< Offset from PTG Base Address: <tt> 0x000C</tt> */
```
Jake-Carter, line 107: Redirect, not delete

Libraries/CMSIS/Device/Maxim/MAX32690/Include/ptg_regs.h
```diff
@@ -147,47 +187,155 @@ typedef struct {
...
 /**@} end of group PTG_RESYNC_Register */
 
 /**
  * @ingroup  ptg_registers
- * @defgroup PTG_INTFL PTG_INTFL
+ * @defgroup PTG_STOP_INTFL PTG_STOP_INTFL
  * @brief    Pulse Train Interrupt Flags
  * @{
  */
-#define MXC_F_PTG_INTFL_PT0_POS                        0 /**< INTFL_PT0 Position */
```
Jake-Carter, line 158: Same comments as other reg files - redirect these but only if the generator can do it.

Libraries/PeriphDrivers/Source/PT/pt_reva.c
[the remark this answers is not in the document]
Jake-Carter: Are you sure?  If `stop_intfl` is not a member of the register struct then won't the compiler throw an error trying to resolve it?  Also, do _all_ parts that use the Rev A PT files have split registers?

This could drive the compiler definition to wrap around the deprecated functions.  Something like `HAS_STOP_REG`?

```C
#ifdef HAS_STOP_REG
ptg->stop_intfl = ptf->stop_intfl;
ptf->ready_intfl = ptf->ready_intfl;
#else
ptg->intfl = ptg->intfl;
#endif
```

```C
#ifdef HAS_STOP_REG
 deprecated("Use MXC_PT_EnableStopInt or MXC_PT_EnableStartInt instead.  See pt.h for more details."))) void MXC_PT_EnableInt(uint32_t pts);
// ... other functions ...
#else
void MXC_PT_EnableInt(uint32_t pts);
#endif
```

Jake-Carter (changes requested):
I agree the split register names are better.  We should not delete the old register, but pass it through.  We should also make sure whatever we end up picking matches the UG exactly.

See comments below for how we might be able to safely deprecate these functions.

sihyung-maxim:
The merge from main resulted in ~3000 file changes to review. The majority of the changed files were from the merge, so the important files to review are in Libraries/CMSIS/Devices/Maxim and in Libraries/PeriphDrivers.

Jake-Carter (changes requested):
I think the clang bot ran off of an outdated script without our updated whitelist.  These formatter changes aren't in the main branch.  I'd suggest reverting the bot format commits, syncing up with main again, then re-running the formatter

Also, did you decide yay or nay on doing the PT updates in new rev implementation file?

Most of my previous comments about the register name deprecation can be ignored.  Doing the register-level passthrough is not worth the effort.  The function deprecation looks good, but I have problems viewing some of it because of the large diffs on the other files.

Libraries/CMSIS/Device/Maxim/MAX32572/Include/ptg_regs.h
```diff
@@ -105,8 +105,8 @@ typedef struct {
  */
 #define MXC_R_PTG_ENABLE                   ((uint32_t)0x00000000UL) /**< Offset from PTG Base Address: <tt> 0x0000</tt> */
 #define MXC_R_PTG_RESYNC                   ((uint32_t)0x00000004UL) /**< Offset from PTG Base Address: <tt> 0x0004</tt> */
-#define MXC_R_PTG_INTFL                    ((uint32_t)0x00000008UL) /**< Offset from PTG Base Address: <tt> 0x0008</tt> */
```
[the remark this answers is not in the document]
sihyung-maxim: Actually, the ME55 doesn't need the direction as it's an unreleased part.

Libraries/CMSIS/Device/Maxim/MAX32572/Include/ptg_regs.h
```diff
@@ -105,8 +105,8 @@ typedef struct {
...
-#define MXC_R_PTG_INTEN                    ((uint32_t)0x0000000CUL) /**< Offset from PTG Base Address: <tt> 0x000C</tt> */
```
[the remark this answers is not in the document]
sihyung-maxim: Ditto^^

"Actually, the ME55 doesn't need the direction as it's an unreleased part."

sihyung-maxim:
Redirecting to new PR: https://github.com/Analog-Devices-MSDK/msdk/pull/337

Reverting did not lower the number of files changed. I created a new branch and PR that's more readable.

PR #285, opened 2022-12-08
UART: Updated WriteCharacter function implementation.

sihyung-maxim:


EdwinFairchild:
I will fix that CI bug when I get back to the office. I ran it locally the PR passes tests.

Jacob-Scheiffler (approved):

Jake-Carter (approved):

Jake-Carter (changes requested):
@sihyung-maxim @Jacob-Scheiffler 

Our drivers don't have an option anymore for a blocking UART write of N bytes without using the full transaction functions that require a request struct.

I propose we add a blocking function with the following header:

```C
int MXC_UART_WriteBytes(mxc_uart_regs_t *uart, const uint8_t *byte, int len)
```

Jacob-Scheiffler:
Good call. I think we should make the function name more explicit though. Maybe MXC_UART_BlockingWrite(...) or something like that?

Jake-Carter:
> Good call. I think we should make the function name more explicit though. Maybe MXC_UART_BlockingWrite(...) or something like that?

I think in general most APIs that offer async options will assume blocking as the default and throw on an `_async`-type designator for the async functions.  We do that for `MXC_UART_Transaction`, `MXC_UART_TransactionAsync`, `MXC_UART_TransactionDMA`, etc.  I like that solution because it makes the common use-case the easy default, and the async option clear.

It looks like `MXC_UART_RevB_WriteCharacter` and `MXC_UART_RevB_WriteCharacterRaw` are exactly the same now as well?...

Jacob-Scheiffler:
1. I think MXC_UART_RevX_WriteChracter should just be removed, because it's not even being used as far as I can tell. All uart_<chip_num>.c files call MXC_UART_Common_WriteCharacter which calls MXC_UART_WriteCharacterRaw once the UART TX FIFO has space. 
2. I see your point about the async naming scheme. If we decide to go that route then we should change MXC_UART_Write to be a blocking function and add MXC_UART_WriteAsync or MXC_UART_WriteNonBlocking. I don't love "MXC_UART_WriteAsync" because it's a bit of a misnomer since it doesn't actually use interrupts.

sihyung-maxim:
Closing this PR. There were issues with this change with UART/printf statements getting cut off, but there probably needs to be an update with our UART implementation.

PR #298, opened 2022-12-13
libphy-update-d3f6d97

kevin-gillespie:
Working making some updates to fix the BLE_FreeRTOS examples.


yc-adi (commented):
Could we remove those log files?

kevin-gillespie:
> Could we remove those log files?

The log files are intended to be left here. Since we can't distribute the source code, this at lease gives us a log of how the library was built.

victor-levi-maxim (approved):

kevin-gillespie:
> > Could we remove those log files?
> 
> The log files are intended to be left here. Since we can't distribute the source code, this at lease gives us a log of how the library was built.

@yc-adi What do you think?

yc-adi (approved):

PR #299, opened 2022-12-14
kws updated with new weights + audio waveform display on TFT, P3.1 wakeup event removed from LP example in Power Down mode

aniktash:


Jake-Carter (approved):

Examples/MAX78000/CNN/kws20_demo/main.c
```diff
@@ -263,6 +275,45 @@ void WUT_IRQHandler()
 }
 #endif
 
+int console_UART_init(uint32_t baud)
```
lorne-maxim, line 278: Doesn't the board file already initialize the console UART?

aniktash: It does with 115k.  This is placed to enable optionally higher baudrate for serial bin transfer.

lorne-maxim: Understood.  Thanks for the clarification.

lorne-maxim (approved):

PR #302, opened 2022-12-15
Sending address with each data packet.

kevin-gillespie:


kevin-gillespie:
This changes the OTA protocol to send a 4 byte address at the start of each data packet. There is additional overhead with this, but it is much more robust.

The current implementation breaks down when the data is delivered to the application layer out of order. We can correct this by using write requests instead of commands, but this drastically reduces the transfer speed.

Sending the address allows us to send write commands, since they can be delivered to the application layer in any order. 

TODO: Update other targets, update phone app, update BLE PyDex

yc-adi:
@EdwinFairchild Recently there are several BLE Examples Test failures.

yc-adi:
Thanks for the update.
I use awk to get port number like:
export devUart3Serial_1=/dev/tty"$(ls -la /dev/serial/by-id | grep -n $DevKitUart3Sn_1 | awk -F tty '{print $2}')".
BTW, what's the link for the PR?

yc-adi (approved):

EdwinFairchild (approved):

kevin-gillespie:
@EdwinFairchild 

We'll have to make sure the various client apps are updated to match this.

PR #307, opened 2022-12-16
(AI85/AI87) Support MXC_LP_EnterSleepMode on RISC-V

Jake-Carter:
All this really does is __WFI() since there is more non-trivial work needed to synchronize the RISC-V core and M4 core sleep mode entry - the details of which are beyond the scope of a PR comment.  The SEMA peripheral needs additional work to expose the required functions to the LP peripheral, and a mailbox signaling protocol needs to be designed to indicate which sleep mode the M4 core should initiate entry into.

This at least gets the latest izer output to build for the RISC-V projects as of https://github.com/MaximIntegratedAI/ai8x-synthesis/pull/273


Jake-Carter:
Let's just get this merged in so the ai8x-synthesis builds at least work for the RISC-V.  A better implementation is on the backlog

sihyung-maxim (approved):

Jacob-Scheiffler (approved):

PR #309, opened 2022-12-19
Ci tests update

EdwinFairchild:
-This will pass once the 32Khz fix is made and merged into main and then main merged into this

### Updates test launcher script
 - script accepts 3 parameters now as follows:
   - device  under test in lower caps  eg:  max32655
   - device serial port serial eg: DT309ZDF
   - CMSIS dap serial : 4440920139230423408230424312233

 - uses board_config.json to find devices instead of hard coded serial numbers
 - for BLE_datc/s connected test there is a short timeout period required to enter then pin
   sometimes if flashing took too long the timeout would pass. Now the script flashes them at the same time
   so that they start up closer to each other and have time to enter the pin.

-

### DIFF
--- a/.github/workflows/BLE_Examples_Test.yml
+++ b/.github/workflows/BLE_Examples_Test.yml
@@ -19,6 +19,7 @@ jobs:
 
     # Steps represent a sequence of tasks that will be executed as part of the job
     steps:
+     # Clean and remove any local modifications
       - name: Clean
         continue-on-error: true
         run: |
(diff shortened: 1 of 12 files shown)


kevin-gillespie:
I reverted the pal_timer changes, but I'm going to make those updates in another PR.

kevin-gillespie (approved):

yc-adi (approved):

PR #312, opened 2022-12-20
SPI RevA: Fixing async status variable overwrite.

Jacob-Scheiffler:
The status variable "async" in spi_reva.c was getting overwritten by other SPI requests. Thus, if the SPI request which set the status variable to "1" had not completed yet, the callback function would not be called. To protect against this, I moved it into the spi_reva_req_state_t struct. This prevents a SPI instance from overwriting another's "async" status.


Jake-Carter (approved):
Looks good!

sihyung-maxim (approved):

PR #323, opened 2023-01-06
Irq attribute

kevin-gillespie:
Adding interrupt attribute to the IRQHandlers.


kevin-gillespie:
@ftariqAnalogTx

Can you see how this works with IAR and Keil? We might have to put the __IRQ on the previous line.

Examples/MAX32690/BLE_FreeRTOS/stack_dats.c
```diff
@@ -190,7 +190,7 @@ static void mainWsfInit(void)
 *  \return None.
 */
 /*************************************************************************************************/
-void WUT0_IRQHandler(void)
+__IRQ void WUT0_IRQHandler(void)
```
ftariqAnalogTx, line 193: I may be able to look into it on Monday, otherwise I will see when I get back.

PR #326, opened 2023-01-10
Matched the RF-PHY-closed full PER ci test.

yc-adi:


Libraries/PeriphDrivers/Source/UART/uart_revb.c
```diff
@@ -843,6 +843,10 @@ int MXC_UART_RevB_AsyncHandler(mxc_uart_revb_regs_t *uart)
     uint32_t numToWrite, numToRead, flags;
     mxc_uart_req_t *req;
 
+    // For PER ci test, need to replace this return value to 0.
```
kevin-gillespie, line 846: We should use sed to comment out the assertion in pal_uart.c as a workaround. I don't think we should modify the driver since we have no reason to believe the error we're receiving isn't real.

yc-adi: There are different cases to return E_INVALID. By my change, I use the sed in ci to only comment out the assertion of what we care.

yc-adi: Okay. I'll modify in the Cordio pal_uart.c.

kevin-gillespie (approved):

EdwinFairchild (approved):

PR #337, opened 2023-01-18
PTG: Added READY and STOP registers for ME12, ME18, and ME55.

sihyung-maxim:
The ME12, ME18, and ME55 will have two more interrupt registers: READY_INTFL (0x18) and READY_INTEN (0x1C). I've renamed the original two interrupt registers: INTFL -> STOP_INTFL (0x08) and INTEN -> STOP_INTEN (0x0C). The ME18 and ME55 are unreleased, and I got the go-ahead to update the ME12 PTG register names.

### DIFF
--- a/Libraries/CMSIS/Device/Maxim/MAX32572/Include/max32572.svd
+++ b/Libraries/CMSIS/Device/Maxim/MAX32572/Include/max32572.svd
@@ -376,6 +376,67 @@
   </peripheral>
 <!--ADC 10-bit Analog to Digital Converter-->
   <peripheral>
+   <name>AESKEYS</name>
+   <description>AES Key Registers.</description>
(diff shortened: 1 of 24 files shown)


sihyung-maxim:
ptg_reva_regs.h will have the new register map as well (READY_ and STOP_ interrupt registers). The register map of the old PT is a subset of the register map of the new PT. The parts with the old IP will not have the new stop/ready functions defined on the chip-specific PT libraries. Also note: the mxc_ptg_regs_t *MXC_PTG instance on the chip-specific level is type-casted with mxc_ptg_reva_regs_t.

Jake-Carter (approved):

Jacob-Scheiffler (approved):

PR #338, opened 2023-01-18
Added dtm_sweep.py

tim-magnusen:
Created DTM sweep test based on conn_sweep.py framework.


yc-adi (approved):

EdwinFairchild (approved):

sihyung-maxim:
The SVD action will fail if the source branch of a PR is from a fork. There's a fix for this merged into main already, but this check isn't important for this PR as there were no register or SVD files that were changed.

kevin-gillespie:
@EricB-ADI @kevin-gillespie

EricB-ADI:
@kevin-gillespie It is effectively the same as conn_sweep_rsl15. @tim-magnusen did you plan on using any non MAX32 boards as a master such as a nordic? If you want I can add an option to use the vendor specific tx test instead of the vendor specific end test, so the nordic boards will still work.

EricB-ADI:
@tim-magnusen The conn_sweep_rsl15 also uses DTM, it was named incorrectly. That will be changed soon. For now, I will merge this so you can use it. I will integrate options to use the end test vendor specific command so you can get more verbose information for receiver testing on MAX32 boards.

PR #371, opened 2023-01-30
Update otas readme

EdwinFairchild:


kevin-gillespie (approved):
I think we accidentally added the peripheral driver library.

PR #372, opened 2023-01-30
Wsf freertos timer

kevin-gillespie:
Fixing a bug where the tail pointer is not updated when removing the tail item.


yc-adi (approved):

EdwinFairchild:
I noticed this only ran BLE test on me17 and examining the yml I realize on me17 has `Libraries/Cordio` in watched files, opps

EdwinFairchild (approved):

PR #383, opened 2023-02-01
Ota improvements

kevin-gillespie:
The goal here is to minimize the code space requirements and improve the reliability. 

In a different branch I've merged BLE_otas with BLE_FreeRTOS.


kevin-gillespie:
Did you set it back to 50 ms?

EdwinFairchild:
> Did you set it back to 50 ms?

yes and erase works fine there, which is where the ci test got hung up. but im getting stuck at the 13th write request.

EdwinFairchild:
I think we need to delay the  write response to the client so it doesn't send packet after packet and were just queuing  up writes. Unfortunately the writes are initiated here:

https://github.com/Analog-Devices-MSDK/msdk/blob/e96ec8f61b47334d01ded783e5224e14f0cd66bc/Libraries/Cordio/ble-host/sources/stack/att/atts_write.c#L162-L168

which is calling this in wdxs_main.c  : 
https://github.com/Analog-Devices-MSDK/msdk/blob/e96ec8f61b47334d01ded783e5224e14f0cd66bc/Libraries/Cordio/ble-profiles/sources/profiles/wdxs/wdxs_main.c#L249-L251

The result of that is `success` and returns to the `atts_write.c` and sends a write response to the client which  then sends the next packet. We could make the response not be success and use ATT_RSP_PENDING, then we send the response after the queue is serviced, but then we are modifying wdxs_main and such..which wont play well with ext flash examples.

EdwinFairchild (approved):

EricB-ADI (approved):

PR #390, opened 2023-02-06
Adding enumeration for btleldoctrl register.

kevin-gillespie:


kevin-gillespie:
We have exiting bit fields, but I wanted to add enumerations for the combined fields. This updated SVD files passes the schema checking tool.

sihyung-maxim (approved):
A lot of whitespace differences in gcr_me17.svd. I commented on the two important changes in that files.

Here are the biggest changes I'll make to the SVD workflow in the coming future:
- Archive generated file differences when SVD workflow fails (similar to BLE_Examples_Test)
- Similar to /clang-format-run, add a comment-triggered command to auto-generate register files, so the user only needs to update the peripheral SVD files (less work for users).

PR #391, opened 2023-02-06
[MSDK-1042]: Resolving hardfault in MAX32672 LP example.

Jacob-Scheiffler:
This PR updates the linkerfile used for the MAX32672 LP example. SRAM 2 and 3 is shutdown in the example so the SRAM usage needs to be contained in SRAM 0 and 1.

### DIFF
--- a/Examples/MAX32672/LP/lp.ld
+++ b/Examples/MAX32672/LP/lp.ld
@@ -32,14 +32,11 @@
  ******************************************************************************/
 
 MEMORY {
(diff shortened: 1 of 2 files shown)


kevin-gillespie (approved):
I would recommend overwriting lp.ld with the default linker file and then changing the SRAM size. That way there is minimal differences between the two.

Jacob-Scheiffler:
> I would recommend overwriting lp.ld with the default linker file and then changing the SRAM size. That way there is minimal differences between the two.

It's done.

Jake-Carter (approved):

PR #403, opened 2023-02-10
Adding workflow to create Cordio platform docs.

AbbyWolf-ADI:


kevin-gillespie (changes requested):
I think we can make this platform agnostic. This documentation will apply to all platforms, not just the MAX32655. 

We should remove the interfaces that we don't really implement. It looks like we've already excluded most of the un-used interfaces, but we can also remove the audio and codec interfaces.

kevin-gillespie:
Can we give some names to the anonymous entries?

Documentation/updateUSERGUIDE-Cordio.py
```diff
@@ -0,0 +1,67 @@
+from pathlib import Path
+from os import listdir
+
```
kevin-gillespie, line 3: So this looks for markdown files and PDFs and adds links to USERGUIDE.md?

Maybe we should move this to the workflows folder?

Documentation/USERGUIDE.md
```diff
@@ -1584,6 +1584,12 @@ The Cordio Bluetooth Low Energy (BLE) library provides a full BLE stack for micr
 - MAX32680
 - MAX32690
 
+#### Cordio Documentation
+
+- [Architecture](ARCHITECTURE.md)
```
kevin-gillespie, line 1589: I think we'll have to add a path for these links to work.

.github/workflows/build-cordio-platform-docs.yml
```diff
@@ -0,0 +1,60 @@
...
+          enable-latex: true
+
+      - name: Build PDF
+        run: |
+          echo "Building PDF..."
+          sudo apt-get install texlive-full
+          sudo make -C Libraries/Cordio/docs/latex
+
+          echo "Copying PDF to Cordio/platform/Documentation"
+          sudo mv Libraries/Cordio/docs/latex/refman.pdf Libraries/Cordio/docs/latex/Cordio-Platform-Documentation.pdf
```
kevin-gillespie, line 47: This filename looks okay, we just need to update what's currently in the PR.

Tools/Bluetooth/README.md
```diff
@@ -0,0 +1,62 @@
+# Bluetooth Tools
+These python tools are used for Bluetooth development and debugging.
+
+## BLE_hci
+Host Controller Interface for controlling Bluetooth Controller devices. Uses a serial port and the HCI interface to send commands and receive events.
+
+```
```
kevin-gillespie, line 7: Can we double check this is the current help text?

kevin-gillespie (changes requested):
Overall this looks good. I'm going to review the README's I moved to make sure we didn't lose significant changes.

Documentation/USERGUIDE.md
AbbyWolf-ADI: tested it a couple different times on my machine, since the workflow copies those file in the same directory, pathing like this works fine

Tools/Bluetooth/README.md
AbbyWolf-ADI: "HCI Tool used to control a Bluetooth Controller through a serial port." is the current help text in the main branch python file, but no README is present there

Documentation/updateUSERGUIDE-Cordio.py
AbbyWolf-ADI: moved, but we will have to do a test to double check the new pathing from that file

Tools/Bluetooth/README.md
kevin-gillespie: If you run with the -h parameter and send the help command then we get this help text.

``` bash
python3 BLE_hci.py -h
python3 BLE_hci.py
help
```

kevin-gillespie:
Can we close this?

PR #405, opened 2023-02-10
Added a simple PER test for ME14/17/18 boards

yc-adi:
Without downloading codes from RF-PHY-closed, do a simple PER test for ME14/17/18 boards with the new Mini-circuits RF switches.

For the 3 boards, it will take about 7 minutes. If need test more points, only need to adjust the configuration file on wall-e in ~/Workspace/ci_config/msdk.json.


EricB-ADI:
I resolved the conflict. It was mad about an extra empty line

PR #407, opened 2023-02-11
Added a simple PER CI test.

yc-adi:


.github/workflows/hard_reset.sh
```diff
@@ -1,23 +0,0 @@
-#!/bin/sh
```
kevin-gillespie, line 1: Why are we deleting these?

.github/workflows/simple_per_test.yml
```diff
@@ -0,0 +1,366 @@
...
+          repository: Analog-Devices-MSDK/msdk
+          ref: main
+
+      - name: clean
+        run: |
+
+          # Clean up the working copy
+          git scorch
+
+          # Update the submodules, this will use ssh
```
kevin-gillespie, line 33: We don't have any submodules anymore.

.github/workflows/simple_per_test.yml
```diff
@@ -0,0 +1,366 @@
...
+            ${MAX32655_STEP}             \
+            2>&1 | tee -a ${CURR_LOG}
+          
+          if [[ $? -ne 0 ]]; then
+            exit 1
+          fi
+          #------------------------------------------------          
+
+      # Unlock even when cancelled or failed
+      - name: unlock_max32655
```
kevin-gillespie, line 196: Where are we doing the locking? We need to make sure that we don't unlock if the lock was unsuccessful.

kevin-gillespie (changes requested):

.github/workflows/simple_per_test.yml
yc-adi: I put all the same operations into a script file so that if there's any change, we only need to change at one place instead of changes everywhere.

.github/workflows/simple_per_test.yml
yc-adi: Got it.

.github/workflows/hard_reset.sh
yc-adi: I moved it. I think it's better to keep all scripts in one place.

.github/workflows/simple_per_test.yml
yc-adi: The locking/unlocking are in the script file. It can handle the case if the test failed or was canceled.

yc-adi:
The new simple_per_test.yml failed because the workflow file is the new one but the script files it uses are not ready in the main branch.

kevin-gillespie (approved):

PR #417, opened 2023-02-14
Cleaning up the clang formatting workflows.

kevin-gillespie:


kevin-gillespie:
@Jake-Carter 

What do you think about removing clang-format-run.yml? This workflow is intended to be a manual dispatch on a branch other than main. We don't use it very often and it doesn't seem necessary, since the formatting checks only run when you create a PR or push to main, and I don't think we want to run this on the main branch directly.

clang-format-run-pr.yml is intended to be run withing a PR. We use this much more frequently.

kevin-gillespie:
I'm having troubling creating something that will allow us to share the REGEX with the linter workflow.

.github/workflows/clang-format-check.yml
```diff
@@ -54,34 +45,7 @@ jobs:
...
-          for change_file in ${CHANGE_FILES}
-          do
-            if [ -f ${change_file} ];
-            then
-            clang-format-${CLANG_VERSION} --verbose --style=file -n -Werror ${change_file}
-            if [ $? != 0 ];
-            then
-              echo ""
-              echo "===================================================="
-              echo "FAIL: ${change_file}"
```
Jake-Carter, line 73: Can we keep this section printing the colored diff and move it into `clang-format-run.sh`?

.github/workflows/clang-format-run.sh
```diff
@@ -1,27 +1,54 @@
...
+else
+  # Accumulate the input arguments into FILES
+  FILES="$*"
+fi
+
+for file in ${FILES}
 do
+  # Determine if we need to make changes
+  set +e
+  clang-format-${CLANG_VERSION} --dry-run --Werror -style=file --verbose ${file}
```
Jake-Carter, line 33: Should probably just use `--verbose` on one of the clang-formats instead of all 3

.github/workflows/clang-format-run.sh
```diff
@@ -1,27 +1,54 @@
...
+  set -e
+
+  if [[ $RETVAL -ne 0 ]]
+  then
+    CHANGES_NEEDED=1
+  fi
+
   # Format the files, this will turn while(1); into while(1)\n;
-  clang-format-${CLANG_VERSION} --verbose -style=file -i ${c_file}
+  clang-format-${CLANG_VERSION} --verbose -style=file -i ${file}
```
Jake-Carter, line 43: `--verbose` here...

.github/workflows/clang-format-run.sh
```diff
@@ -1,27 +1,54 @@
...
 
   # Remove single line ';' and replace with "{}"
-  perl -i -pe 's/\s+;\s/{}\n/' ${c_file}
+  perl -i -pe 's/\s+;\s/{}\n/' ${file}
 
   # Re-format the files
-  clang-format-${CLANG_VERSION} --verbose -style=file -i ${c_file}
+  clang-format-${CLANG_VERSION} --verbose -style=file -i ${file}
```
Jake-Carter, line 49: `--verbose` here...

.github/workflows/clang-format-run.sh
Jake-Carter, line 43: Move the actual formatting work inside...

```bash
if [[ $RETVAL -ne 0 ]]
then
    CHANGES_NEEDED=1
    # Here!
fi
```

To avoid triplification of checks and reduce script time.

Jake-Carter (changes requested):

Jake-Carter:
> @Jake-Carter
> 
> What do you think about removing clang-format-run.yml? This workflow is intended to be a manual dispatch on a branch other than main. We don't use it very often and it doesn't seem necessary, since the formatting checks only run when you create a PR or push to main, and I don't think we want to run this on the main branch directly.
> 
> clang-format-run-pr.yml is intended to be run withing a PR. We use this much more frequently.

I think we can remove it.  If I need to manually format my branch I will run the formatter locally.

.github/workflows/clang-format-check.yml
kevin-gillespie: I was wondering where the colors came from O_O

Jake-Carter (approved):
Nice, she runs 300% faster and the output is pretty

EdwinFairchild (approved):

PR #420, opened 2023-02-14
Fix AI85/AI87 Image Capture Examples (ImgCapture, CSI2)

Jake-Carter:
As of https://github.com/Analog-Devices-MSDK/msdk/pull/283 UART writes became non-blocking, which broke the serial console used by ImgCapture and CSI2 on the AI85/AI87.

This PR re-introduces a blocking write function into the example's local source code until a long-term solution in https://github.com/Analog-Devices-MSDK/msdk/pull/285 is implemented.

It also applies a temporary workaround to some newly discovered CSI2 issues to get captures working reliably again.  (CSI2 output buffer memory contention, see AI87-123, AI87-129).

These are fundamental fixes that _must_ be released in the next MSDK version, but also contain the foundations for longer-term solutions that will be released by the grace of the microcontroller gods.

Related Jira:  MSDK-917


Examples/MAX78002/CSI2/main.c
```diff
@@ -195,7 +210,7 @@ void service_console()
     clear_serial_buffer();
 }
 
-volatile int buttonPressed = 0;
```
Jacob-Scheiffler, line 198: Be careful taking out the volatile, may cause example to hang if compiler optimization level is increased.

Jacob-Scheiffler (approved):

sihyung-maxim (approved):

Examples/MAX78002/CSI2/main.c
Jake-Carter: Thanks, good catch

PR #426, opened 2023-02-16
REV A ADC Driver Fix

EricB-ADI:
Data returned from the following function is not correct. 

https://github.com/Analog-Devices-MSDK/msdk/blob/27e4f783a8c5ea709aabad22633a3a4e1f6dd097/Libraries/PeriphDrivers/Source/ADC/adc_reva.c#L328

The cast shown below causes the format of the value to be messed up.

https://github.com/Analog-Devices-MSDK/msdk/blob/27e4f783a8c5ea709aabad22633a3a4e1f6dd097/Libraries/PeriphDrivers/Source/ADC/adc_reva.c#L355

Instead of declaring the data as an int, I swiched to uint16_t to avoid this issue. Casting it as an int upon return fixes this issue and the data matches, the data coming from _MXC_ADC_GetData_

I haven't looked at the other revisions of the driver, but it is probably in them as well. 

The example doesn't catch this issue, because MXC_ADC_GetData is used directly.

### DIFF
--- a/Libraries/PeriphDrivers/Source/ADC/adc_reva.c
+++ b/Libraries/PeriphDrivers/Source/ADC/adc_reva.c
@@ -327,7 +327,9 @@ void MXC_ADC_RevA_DisableMonitorAsync(mxc_adc_monitor_t monitor)
 
 int MXC_ADC_RevA_StartConversion(mxc_adc_reva_regs_t *adc, mxc_adc_chsel_t channel)
 {
-    int data;
+    uint16_t data;
+    
+    
     int error;
 #if TARGET_NUM != 32650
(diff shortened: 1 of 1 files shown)


kevin-gillespie:
So all of the error codes are negative and what we get in the successful case is a raw unsigned value?

EricB-ADI:
Yeah the interface is problematic to begin with, but the actual read is now correct when it is returned. It would be better to just pass a pointer to the data so the error code is separate, but I don't know how much stuff is relying on this.

Jake-Carter (approved):
Looks good!

kevin-gillespie (approved):

PR #428, opened 2023-02-17
Minor correction in Documentation

viviand0097:
The LED blinks at 1Hz, because a full cycle takes 1 second


Jake-Carter (approved):
Thanks @viviand0097, you're right.  The LED _toggles_ at 2Hz, but that gets us a _blink_ at 1Hz.  Looks like this got copy/pasted to most of the Hello World readmes.

PR #436, opened 2023-02-22
Script to plot dtm test results

EricB-ADI:
This script is used to plot the results of the DTM sweep in bluetooth tools


EdwinFairchild (approved):

EdwinFairchild:
I should probably change the order of operations in yaml for BLE tests, it tried to lock the boards and timed-out thus failing, if it had successfully locked that boards it would then concluded it does not need to run BLE test on this and passed. 
So i should only proceed to lock step after it decides a test is warranted.

EricB-ADI:
I agree

AbbyWolf-ADI (approved):

PR #440, opened 2023-02-23
Update call to lock boards

EdwinFairchild:


yc-adi (commented):
Is it possible to check if the input is with ".txt", then it will follow the old way?

EricB-ADI (approved):

EricB-ADI:
> Is it possible to check if the input is with ".txt", then it will follow the old way?

I thought we just wanted to ask for the board name?

EdwinFairchild:
> Is it possible to check if the input is with ".txt", then it will follow the old way?

Resource_Share.py has been modified to accept a board name and it will look for that board name in the json file and make a `.txt` file given the "lockfile" property.

kevin-gillespie:
I added something like this to trim exiting file extensions and paths. The name will still have to match what's in the JSON file.

```python
#! /usr/bin/env python3

from pathlib import Path

board0="/home/kgills/max32655_W1.txt"
board1="max32655_W1.txt"
board2="max32655_W1"

print(Path(board0).resolve().stem)
print(Path(board1).resolve().stem)
print(Path(board2).resolve().stem)
```
```bash
max32655_W1
max32655_W1
max32655_W1
```

EdwinFairchild:
> I added something like this to trim exiting file extensions and paths. The name will still have to match what's in the JSON file.
> 
> ```python
> #! /usr/bin/env python3
> 
> from pathlib import Path
> 
> board0="/home/kgills/max32655_W1.txt"
> board1="max32655_W1.txt"
> board2="max32655_W1"
> 
> print(Path(board0).resolve().stem)
> print(Path(board1).resolve().stem)
> print(Path(board2).resolve().stem)
> ```
> 
> ```shell
> max32655_W1
> max32655_W1
> max32655_W1
> ```

Is this referring to the paths we use in the bash scripts to call Resource_share.py ? otherwise, I dont follow.

kevin-gillespie:
Yes, I modified Resource_share.py to trim the board input.

yc-adi:
In this PR (also in next RF-PHY-closed PR), all inputs of Resource_Share.py are changed to board name instead of board<n>.txt.

yc-adi (approved):

PR #454, opened 2023-02-27
TMR RevB: Implemented Set/Get ClockSourceFreq functions.

sihyung-maxim:


Jacob-Scheiffler (approved):

Libraries/PeriphDrivers/Include/MAX32655/tmr.h
```diff
@@ -154,6 +154,23 @@ typedef void (*mxc_tmr_complete_t)(int error);
  */
 int MXC_TMR_Init(mxc_tmr_regs_t *tmr, mxc_tmr_cfg_t *cfg, bool init_pins);
 
+/**
+ * @brief   Save the TMRn's Clock Source Frequency.
+ * @param   tmr          Pointer to timer module to save clock source frequency.
+ * @param   clksrc_freq  Clock Source Frequency (Hz) value to save.
+ */
+void MXC_TMR_SetClockSourceFreq(mxc_tmr_regs_t *tmr, int clksrc_freq);
```
Jake-Carter, line 162: I think we could improve this wording.  Something like:

```C
/**
 * @brief   Set the Clock Source Frequency for the specified Timer (TMR) instance.
 * @param   tmr          Timer instance to update.
 * @param   clksrc_freq  Target Clock Source Frequency (Hz).
 */
```

Libraries/PeriphDrivers/Source/TMR/tmr_revb.c
```diff
@@ -443,7 +443,7 @@ int MXC_TMR_RevB_GetTime(mxc_tmr_revb_regs_t *tmr, uint32_t ticks, uint32_t *tim
     uint32_t prescale = (tmr->ctrl0 & MXC_F_TMR_REVB_CTRL0_CLKDIV_A) >>
                         MXC_F_TMR_REVB_CTRL0_CLKDIV_A_POS;
 
-    uint32_t timerClock = MXC_TMR_RevB_GetClockSourceFreq(tmr);
+    int timerClock = MXC_TMR_RevB_GetClockSourceFreq(tmr);
```
Jake-Carter, line 446: Make this an `unsigned int` since we cannot have negative clock frequencies.  This will change the return type of all the getters as well.

Libraries/PeriphDrivers/Include/MAX32655/tmr.h
Jake-Carter, line 162: This should also not be a publicly accessible API function unless it's updated to actually set the clock source as well.  In which case it should become `MXC_TMR_SetClockSource`

If left as its current implementation move this inside of the C files.

Jake-Carter (changes requested):

Libraries/PeriphDrivers/Source/TMR/tmr_revb.c
Jake-Carter: Nevermind - we check this for error codes.

Just update the function definitions to use `unsigned int`

Jake-Carter: Nevermind - we use the negative codes for errors

Jake-Carter (approved):

PR #492, opened 2023-03-15
Cordio lib

kevin-gillespie:
Building a Cordio library. I used the peripheral driver layout as a reference. 

TODO: Select files that need to be re-built for each application.


Libraries/Cordio/wsf/sources/targets/freertos/wsf_trace.c
```diff
@@ -218,7 +218,31 @@ void WsfTrace(const char *pStr, ...)
     }
   }
 }
-#endif
+#if WSF_TRACE_ENABLED_VERBOSE == TRUE
```
kevin-gillespie, line 221: This will allow the build to pass, but will cause problems when the trace options change. Still not sure which outcome is better. 

**TRACE** options could require a re-build between applications. If the library is built with **TRACE=0** it will be optimized for code size and remove all of the trace messaging. Adjacent applications will still build and link with **TRACE=1**, but the trace messages will not be printed until the library is re-built.

If the library is built with **TRACE=2** and adjacent applications build with **TRACE=0**, the messages will not be printed, but the text will remain in the code space until the library is rebuilt with the appropriate option.

Libraries/Cordio/ble-apps/build/common/gcc/config.mk
```diff
@@ -66,14 +66,6 @@ ifneq ($(DEBUG),0)
 CFG_DEV         += WSF_BUF_STATS=1
 CFG_DEV         += WSF_ASSERT_ENABLED=1
 endif
-ifeq ($(TRACE),1)
```
kevin-gillespie, line 69: TODO: Add verbose settings here.

Jake-Carter (approved):
@kevin-gillespie From a build system perspective I think this looks good and is a major improvement.

Since `TRACE` affects the build significantly, I made a small tweak so the different trace levels build as their own variants.  See https://github.com/Analog-Devices-MSDK/msdk/pull/492/commits/7c8efb4d02394f8a56623cf58eb39a40d0572d58

This makes TRACE changes transparent to the user and removes some documentation overhead.  `project.mk` can be updated to remove the comment about rebuilding Cordio on TRACE changes.  What do you think?

EdwinFairchild (approved):
this does not require anything new to passed on to CI test right? since everything is built from scratch.

kevin-gillespie:
> this does not require anything new to passed on to CI test right? since everything is built from scratch.

Correct. We shouldn't have to modify the CI.

kevin-gillespie:
> @kevin-gillespie From a build system perspective I think this looks good and is a major improvement.
> 
> Since `TRACE` affects the build significantly, I made a small tweak so the different trace levels build as their own variants. See [7c8efb4](https://github.com/Analog-Devices-MSDK/msdk/commit/7c8efb4d02394f8a56623cf58eb39a40d0572d58)
> 
> This makes TRACE changes transparent to the user and removes some documentation overhead. `project.mk` can be updated to remove the comment about rebuilding Cordio on TRACE changes. What do you think?

I thought about this and I think it will work fine. My concern is that the variants will grow out of control if we're not careful.

I will update the comments and docs to reflect this.

yc-adi:
@kevin-gillespie 
https://github.com/Analog-Devices-MSDK/msdk/blob/e983cefa5841af159364aeec085cbdb1223d354c/Examples/MAX32655/BLE5_ctr/project.mk#L20
"TRACE = 2" caused the BLE TIFS test fail.

PR #496, opened 2023-03-17
Added PER mask

yc-adi:


EricB-ADI (approved):

Tools/Bluetooth/conn_sweep.py
```diff
@@ -427,89 +320,115 @@ def run_script_reset_board(sh_file):
...
+                mini_RCDAT = mini_RCDAT_USB(Namespace(atten=set_val))
+            
+            sleep(0.1)
+
+            print("\nReset the packet stats.")
+            hciSlave.cmdFunc(Namespace(cmd="0102FF00"), timeout=10.0)
+            hciMaster.cmdFunc(Namespace(cmd="0102FF00"), timeout=10.0)
+
+            print("\nSlave listenFunc")
+            hciSlave.listenFunc(Namespace(time=1, stats="False"))
```
kevin-gillespie, line 342: I think we can remove these two listen calls. I'm pretty sure the intended effect is to clear out any pending events before checking the packet statistics.

kevin-gillespie: Since we're doing this on lines 351 and 352, I think these 2 are redundant.

kevin-gillespie (changes requested):

Tools/Bluetooth/conn_sweep.py
yc-adi: Okay, I'll remove them. Is there a easy way to get all the HCI commands used by CMW?

kevin-gillespie: Not that I can think of. We can use the TRACE output while it's being tested to see what commands are sent. It's something along the lines of these sequences. 

```
HCI Reset
TxTest
EndTest
HCI Reset

HCI Reset
RxTest
EndTest
HCI Reset
```

kevin-gillespie (approved):

EricB-ADI (approved):

PR #501, opened 2023-03-21
feat(Boards): Update LP board settings for MAX32655 RevB1

kevin-gillespie:
Updating MAX32655 LP settings for revision B1.

### DIFF
--- a/Examples/MAX32655/BLE_FreeRTOS/main.c
+++ b/Examples/MAX32655/BLE_FreeRTOS/main.c
@@ -124,8 +124,6 @@ int main(void)
     volatile int i;
     for (i = 0; i < 0x3FFFFF; i++) {}
 
-    GPIO_PrepForSleep();
-
     /* Start the BLE application */
     bleStartup();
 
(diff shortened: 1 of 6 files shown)


kevin-gillespie:
Ok I modified that function to ```GPIO_Init()``` that is called at Board_Init. 

@Jacob-Scheiffler The VDDIO/VDDIOH settings still don't seem correct. I'm measuring ~20 uA through VREGI when we're in standby mode. I think this should be close to 0.

lorne-maxim (approved):

Jacob-Scheiffler:
> TODO: We'll have to set the VCOREB voltage to 0.81 whenever we enter standby/backup mode. We will probably have to raise the voltage back once we exit standby/backup.

You are correct. I was seeing Standby measurements around 5uA, although it varied slightly depending on VCOREB/VREGO_B settings.

kevin-gillespie:
Tom suggested we look at the RTC oscctrl register. 

MXC_RTC->oscctrl = 0x9, but should be 0 to minimize RTC current consumption.

Jake-Carter:
@lorne-maxim @Jacob-Scheiffler any idea what the status of this one is?  Good to merge or do we need to take a deeper look at the MXC_RTC->oscctrl setting?

Jacob-Scheiffler:
@jake-carter No idea where Kevin left off with this. I haven't heard anything from design about any further LP testing for the ME17B either.

PR #504, opened 2023-03-22
implements script to mass erase devices upon unlock

EdwinFairchild:
-check to see which boards needs testing
-if at least one needs testing it locks all 3 because its safer and a lot of issues arose when only locking one and waiting for others
-test all boards
-And lastly does a mass erase of any locked boards before unlocking. **Does this at the yaml level so if test script is interrupted because of a canceled workflow the board will still be left in a known state.**

### DIFF
--- a/.github/workflows/scripts/mass_erase_board.sh
+++ b/.github/workflows/scripts/mass_erase_board.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+#script accepts 2 arguments: device , device key in boards config
+
+#****************************************************************************************************
+# Function accepts parameters:device , CMSIS-DAP serial #
+function erase_with_openocd() {
+    echo "-----------------------------------------------------------------------------------------"
+    printf "> Erasing $1 : $2 \r\n"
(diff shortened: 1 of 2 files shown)


yc-adi (approved):

.github/workflows/scripts/mass_erase_board.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+#script accepts 2 arguments: device , device key in boards config
+
+#****************************************************************************************************
+# Function accepts parameters:device , CMSIS-DAP serial #
+function erase_with_openocd() {
```
kevin-gillespie, line 6: We could probably also erase the external flash devices with OpenOCD. The mass erase would take too long, but I think we can erase 1 or 2 sectors pretty quickly.

kevin-gillespie: Well maybe not :/ 

The MAX32655 doesn't have the XIP SPI, so the SPI erase won't work with OpenOCD.

kevin-gillespie (approved):

PR #517, opened 2023-03-28
Cordio:  Initialize WSF Message Buffers with Null Terminator

EdwinFairchild:
If you send a short string after sending a long one , the remains of the long one still print and sometimes junk. 
Seems like the other side is not sending a terminating `\0`
Found this 11 months ago. The time has come.


Jake-Carter:
@EdwinFairchild status on this one?

EdwinFairchild:
@yc-adi @AbbyWolf-ADI can you guys give this a glance it simply terminates a string before we print it, because sometimes we print junk

yc-adi (approved):

Jake-Carter (changes requested):
A better solution for this would be to initialize the entire buffer with `\0` when you allocate it in `WsfMsgAlloc`

https://github.com/Analog-Devices-MSDK/msdk/blob/80e908b1ba2464e0dd62b449ba82b7feede6566e/Libraries/Cordio/wsf/sources/targets/baremetal/wsf_msg.c#L68

Ex:
```C
  pMsg = WsfBufAlloc(len + sizeof(wsfMsg_t));
  memset(pMsg, '\0', len + sizeof(wsfMsg_t));
```

EdwinFairchild:
I can dig it!

Jake-Carter (commented):
@EdwinFairchild 
Don't forget about the freertos file too :)

https://github.com/Analog-Devices-MSDK/msdk/blob/9bf68cf780920cc3d9654418c5722543f2e69c89/Libraries/Cordio/wsf/sources/targets/freertos/wsf_msg.c#L67

PR #519, opened 2023-03-28
Adding Makefile targets for flashing.

kevin-gillespie:
This will allow us to ```make flash.openocd```.

I added another target for JLink ```make flash.jlink```. This isn't a part of our SDK, but I know that some of our customers prefer this to OpenOCD. 

What do we think of the file path and include in the Makefile?

### DIFF
--- a/Examples/MAX32655/Hello_World/Makefile
+++ b/Examples/MAX32655/Hello_World/Makefile
@@ -406,6 +406,3 @@ clean:
 
 # The rule to clean out all the build products.
 distclean: clean libclean
-
-# Include rules for flashing
-include $(MAXIM_PATH)/Tools/Flash/flash.mk
--- a/Examples/MAX32655/Hello_World/Makefile
+++ b/Examples/MAX32655/Hello_World/Makefile
@@ -406,3 +406,6 @@ clean:
 
 # The rule to clean out all the build products.
 distclean: clean libclean
+
+# Include rules for flashing
+include $(MAXIM_PATH)/Tools/Flash/flash.mk
--- a/Libraries/CMSIS/Device/Maxim/MAX32520/Source/GCC/max32520.mk
+++ b/Libraries/CMSIS/Device/Maxim/MAX32520/Source/GCC/max32520.mk
@@ -88,4 +88,4 @@ LIBPATH+=$(CMSIS_ROOT)/Device/Maxim/$(TARGET_UC)/Source/GCC
 include $(CMSIS_ROOT)/Device/Maxim/$(TARGET_UC)/Source/GCC/gcc.mk
 
 # Include rules for flashing
-include $(MAXIM_PATH)/Tools/Flash/flash.mk
+include $(CMSIS_ROOT)/../../Tools/Flash/flash.mk
--- a/Libraries/CMSIS/Device/Maxim/MAX32520/Source/GCC/max32520.mk
+++ b/Libraries/CMSIS/Device/Maxim/MAX32520/Source/GCC/max32520.mk
@@ -86,3 +86,6 @@ LIBPATH+=$(CMSIS_ROOT)/Device/Maxim/$(TARGET_UC)/Source/GCC
 
 # Include the rules and goals for building
 include $(CMSIS_ROOT)/Device/Maxim/$(TARGET_UC)/Source/GCC/gcc.mk
+
+# Include rules for flashing
+include $(MAXIM_PATH)/Tools/Flash/flash.mk
--- a/Tools/Flash/flash.mk
+++ b/Tools/Flash/flash.mk
@@ -1,0 +1,81 @@
+################################################################################
+ # Copyright (C) 2023 Maxim Integrated Products, Inc., All Rights Reserved.
+ #
+ # Permission is hereby granted, free of charge, to any person obtaining a
+ # copy of this software and associated documentation files (the "Software"),
+ # to deal in the Software without restriction, including without limitation
+ # the rights to use, copy, modify, merge, publish, distribute, sublicense,
+ # and/or sell copies of the Software, and to permit persons to whom the
+ # Software is furnished to do so, subject to the following conditions:
+ #
+ # The above copyright notice and this permission notice shall be included
+ # in all copies or substantial portions of the Software.
+ #
+ # THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+ # OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+ # MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.
+ # IN NO EVENT SHALL MAXIM INTEGRATED BE LIABLE FOR ANY CLAIM, DAMAGES
+ # OR OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE,
+ # ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+ # OTHER DEALINGS IN THE SOFTWARE.
+ #
+ # Except as contained in this notice, the name of Maxim Integrated
+ # Products, Inc. shall not be used except as stated in the Maxim Integrated
+ # Products, Inc. Branding Policy.
+ #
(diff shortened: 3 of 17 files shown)


Tools/Flash/flash.mk
```diff
@@ -0,0 +1,81 @@
...
+# ADAPTER_SN=04090000068682d300000000000000000000000097969906
+
+# Location of the OpenOCD install directory, tcl or scripts folder
+OPENOCD_SCRIPTS ?= ${MAXIM_PATH}/Tools/OpenOCD/scripts
+
+# Update with the connected debugger
+OPENOCD_ADAPTER ?= cmsis-dap.cfg
+# OPENOCD_ADAPTER ?= jlink.cfg
+
+HEX_FILE        := ${BUILD_DIR}/${PROJECT}.hex
```
Jake-Carter, line 53: Hex files don't get built unless we target `make release`.  Any special reason for using them instead of the elf?

Examples/MAX32655/Hello_World/Makefile
```diff
@@ -406,3 +406,6 @@ clean:
 
 # The rule to clean out all the build products.
 distclean: clean libclean
+
+# Include rules for flashing
+include $(MAXIM_PATH)/Tools/Flash/flash.mk
```
Jake-Carter, line 411: Let's include this at the bottom of gcc.mk so we don't have to regenerate the project Makefiles

Tools/Flash/flash.mk
```diff
@@ -0,0 +1,81 @@
...
+
+ifeq ($(OS),Windows_NT)
```
Jake-Carter, line 55: Where does OS come from?  We should probably explicitly check for MSYS to guarantee the availability of `cygpath`.

Ex:
https://github.com/Analog-Devices-MSDK/msdk/blob/6c03d9c9397812b6d90835c9446fded5685eca25/Libraries/CMSIS/Device/Maxim/MAX78000/Source/GCC/gcc.mk#L100

Jake-Carter (changes requested):
This'll be awesome.  A few questions below

Tools/Flash/flash.mk
kevin-gillespie: The hex fill will be created with the dependencies in the make target.

We've also seen issues with the OpenOCD program command and .elf files. If the starting address is offset, it will erase the first pages of flash up to the starting address.

Tools/Flash/flash.mk
kevin-gillespie: I'm not sure where the OS variable comes from. I agree that this doesn't guarantee the availability of cygpath.

Examples/MAX32655/Hello_World/Makefile
kevin-gillespie: What about in the board Makefiles? This could be used with other compilers.

Tools/Flash/flash.mk
kevin-gillespie: Modified this to check for cygpath availability. Not sure how much conversion we'll have to do once we start using native Make on Windows.

Tools/Flash/flash.mk
```diff
@@ -0,0 +1,107 @@
...
+OPENOCD_ADAPTER ?= cmsis-dap.cfg
+# OPENOCD_ADAPTER ?= jlink.cfg
+
+HEX_FILE        := ${BUILD_DIR}/${PROJECT}.hex
+HEX_FILE_PATH   := ${HEX_FILE}
+ifeq ($(OS),Windows_NT)
+JLINKEXE        := JLink.exe
+ECHO            := echo -e
+OPENOCDEXE      := openocd.exe
+
```
lorne-maxim, line 64: Should the first setting of JLINKEXE, ECHO, and OPENOCDEXE use '?=" instead of ":=" so users can set their own values of these if they want?  For example, if they don't have jlink in their path, they could set this value to a full path.

lorne-maxim (approved):
One optional change (see in-line comment).

Tools/Flash/flash.mk
kevin-gillespie: Good point. I made this change.

Libraries/CMSIS/Device/Maxim/MAX32520/Source/GCC/max32520.mk
```diff
@@ -86,3 +86,6 @@ LIBPATH+=$(CMSIS_ROOT)/Device/Maxim/$(TARGET_UC)/Source/GCC
 
 # Include the rules and goals for building
 include $(CMSIS_ROOT)/Device/Maxim/$(TARGET_UC)/Source/GCC/gcc.mk
+
+# Include rules for flashing
+include $(CMSIS_ROOT)/../../Tools/Flash/flash.mk
```
Jake-Carter, line 91: I was going to yell at you for using `CMSIS_ROOT` instead of `MAXIM_PATH` but I see that the rest of the file is the real problem so you're off the hook.

Tools/Flash/flash.mk
```diff
@@ -0,0 +1,107 @@
...
+# Example for CMSIS-DAP: 04090000068682d300000000000000000000000097969906
+#
+# ADAPTER_SN=229002239
+# ADAPTER_SN=04090000068682d300000000000000000000000097969906
+
+# Location of the OpenOCD install directory, tcl or scripts folder
+
+OPENOCD_SCRIPTS_DEFAULT = ${MAXIM_PATH}/Tools/OpenOCD/scripts
+ifeq ($(OPENOCD_SCRIPTS),)
+$(warning Warning: OPENOCD_SCRIPTS undefined, using ${OPENOCD_SCRIPTS_DEFAULT})
```
Jake-Carter, line 50: Can we get rid of this warning?  It clutters the output a little and 99% of the time we will use the default path

Jake-Carter (commented):
Looks good.  I suggest we remove the `OPENOCD_SCRIPTS` warning, but other than that I don't see any other changes to make.

Tested on my MAX78002EVKIT.  Works great with standard projects.  RISC-V projects don't work for some reason...

If I run `make` from `MAX78002/CNN/imagenet-riscv` it includes the flash file correctly:

```bash
C:\Users\JCarter3\repos\fork\msdk\Examples\MAX78002\CNN\imagenet-riscv>make
make[1]: Entering directory '/c/Users/JCarter3/repos/fork/msdk/Examples/MAX78002/CNN/imagenet-riscv'
here!
/C/Users/JCarter3/repos/fork/msdk/Libraries/CMSIS/../../Tools/Flash/flash.mk:50: Warning: OPENOCD_SCRIPTS undefined, using C:/Users/JCarter3/repos/fork/msdk/Tools/OpenOCD/scripts
make[1]: Leaving directory '/c/Users/JCarter3/repos/fork/msdk/Examples/MAX78002/CNN/imagenet-riscv'
make[1]: Entering directory '/c/Users/JCarter3/repos/fork/msdk/Examples/MAX78002/CNN/imagenet-riscv'
/bin/sh: line 1: gcc: command not found
expr: syntax error: unexpected argument ‘4’
here!
/C/Users/JCarter3/repos/fork/msdk/Libraries/CMSIS/../../Tools/Flash/flash.mk:50: Warning: OPENOCD_SCRIPTS undefined, using C:/Users/JCarter3/repos/fork/msdk/Tools/OpenOCD/scripts
/C/Users/JCarter3/repos/fork/msdk/Libraries/CMSIS/../../Tools/Flash/flash.mk:84: Warning: cygpath unavailable to convert path name 
to Windows format
make[1]: Leaving directory '/c/Users/JCarter3/repos/fork/msdk/Examples/MAX78002/CNN/imagenet-riscv'
```

However if I then run `make flash.openocd`:

```bash
C:\Users\JCarter3\repos\fork\msdk\Examples\MAX78002\CNN\imagenet-riscv>make flash.openocd
make: *** No rule to make target 'flash.openocd'.  Stop.
```

Not sure why this is failing

kevin-gillespie:
Just to follow up on the build failure, this is an issue with that particular example because all of the available build targets are explicitly set in the Makefile. It doesn't include any other Makefiles.

Jake-Carter (approved):

Jake-Carter:
> Just to follow up on the build failure, this is an issue with that particular example because all of the available build targets are explicitly set in the Makefile. It doesn't include any other Makefiles.

TODO:  Improve RISC-V Makefile system for AI85/AI87 to match what we did for the BLE micros

PR #520, opened 2023-03-28
Erase boards  to start from known state

EdwinFairchild:
Since these boards are now shared with more workflows than before, it could be a failed workflow leaves the boards in some unknown state so I am going to also erase  them  on start of test and end of test which was previously merged.


EdwinFairchild:
@EricB-ADI @yc-adi guys can I get a review on this
The boards are erase at end and start of the script just to always have a known state. I know its a lot but takes less than 5 seconds to erase all boards at once.

yc-adi (approved):
Yes, I agree with it. I'll follow that in the timing test and PER test.

AbbyWolf-ADI (approved):

PR #524, opened 2023-03-29
Building archives with text file arguments.

kevin-gillespie:
This will address the same issue we had with linking the .elf files. The command line list of argmuents is too long on Windows machines. This will echo those arguments to a text file and use that file as the input the the archiver.


tim-magnusen (approved):

Jake-Carter (approved):
We should add Windows runners to our CI to catch these things

kevin-gillespie:
> We should add Windows runners to our CI to catch these things

Good idea. It looks like we can use GitHub hosted windows machines, they're just billed at a higher rate than the Ubuntu machines. Our current consumption isn't close to the monthly limit, so adding a Windows build test shouldn't matter.

PR #529, opened 2023-03-30
Modifying MAX32672 LP example to get lowest current measurements.

Jacob-Scheiffler:
This PR adds instructions in the MAX32672 LP example README on how to configure the EV kit to achieve low power measurements that match what is listed in the datasheet. Additionally, a function was added in the example code that sets all GPIOs low except for the SWD pins, and the push button/UART pins if they are in use.

### DIFF
--- a/Examples/MAX32672/LP/main.c
+++ b/Examples/MAX32672/LP/main.c
@@ -204,7 +204,8 @@ void configure_gpios(void)
 
 int main(void)
 {
-    // Delay to prevent bricks
+    // Delay to provide the debugger with a window to connect.
+    // Low-power modes shut down SWD
     MXC_Delay(MXC_DELAY_SEC(2));
 
     PRINT("\n************ Low Power Mode Example ************\n\n");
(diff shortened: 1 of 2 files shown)


Examples/MAX32672/LP/main.c
```diff
@@ -163,8 +165,49 @@ void setTrigger(int waitForTrigger)
...
+    out_set.func = MXC_GPIO_FUNC_OUT;
+    out_set.pad = MXC_GPIO_PAD_NONE;
+    out_set.vssel = MXC_GPIO_VSSEL_VDDIOH;
+    MXC_GPIO_Config(&out_set);
+    MXC_GPIO_OutSet(out_set.port, out_set.mask);
+}
+
 int main(void)
 {
+    // Delay to provide the debugger with a window to connect.
```
Jake-Carter, line 207: FYI updated this comment.

In general I have been leaning on the approach of "gating" the application start behind a push-button press or serial connection.  "Push PB1 to start...", etc.  The delay is ol' reliable so I'll leave it up to you

Jake-Carter (approved):

Examples/MAX32672/LP/main.c
sihyung-maxim: ^^^^

sihyung-maxim (approved):

PR #521, opened 2023-03-28
MXC_SYS:  Add Support for External Clock Inputs, Fix MXC_SYS API Documentation

Jake-Carter:
Resolves #518 

## Overvew

- Enable the external clock input as an option for the system clock source for microcontrollers that support it (mxc_sys).
- Add the `EXT_CLK` example for the MAX32670.
- Update Doxygen files to include static members.  This fixes missing documentation for some mxc_sys functions (`MXC_SYS_Crit_xxx`) and macro (`MXC_CRITICAL(xxx)`).
- Fix missing mxc_sys section in API doxygen for all micros.
- Improve mxc_sys API documentation (see [here ](https://jake-carter.github.io/msdk/Libraries/PeriphDrivers/Documentation/MAX32670/group__mxc__sys.html) for example)

## Testing

- Tested with the MAX32670EVKIT (Rev B) with a 2Mhz 50% duty cycle square wave provided from a function generator.  At higher clock frequencies (i.e. 20Mhz) operation became unreliable due to ringing/impedance mismatch causing degradation of the clock signal.  This was a limitation of the test setup.  Given a clean signal, the microcontroller is expected to function up to the specs of its datasheet.

PR #534, opened 2023-03-31
MAX32672: Updated AESKEYS and SYS Registers, Deprecated GPIO SetVSSEL and Pad Options.

sihyung-maxim:
Changes:
- Added two AES Keys (System and User) to ME21.
- Fixed SVD Workflow issues.
- Deprecated GPIO_SetVSSEL and pad options for ME21.
- Updated all register files with Doxygen additions.


sihyung-maxim:
Making a change where all my future PRs will have the title with the chip names instead of the die names.

Jacob-Scheiffler (commented):
Looks good to me. Probably should get some other approvals on this since I made some of these.

cwarer1 (approved):

Libraries/CMSIS/Device/Maxim/MAX32672/Include/aeskeys_regs.h
```diff
@@ -84,7 +86,11 @@ extern "C" {
 /**
  * @ingroup aeskeys_registers
```
Jake-Carter, line 87: I wonder if we can also clean up our Doxygen output for deprecated functions/structs.

Primarily we can add [@deprecated](https://www.doxygen.nl/manual/commands.html#cmddeprecated) to the header for the structs and register modules.

I'm also not sure exactly how the [current Doxygen deprecated list](https://analog-devices-msdk.github.io/msdk/Libraries/PeriphDrivers/Documentation/MAX32672/deprecated.html) is being generated.  There's not a `@deprecated` tag anywhere in the file.

Libraries/CMSIS/Device/Maxim/MAX32672/Include/max32672.h
```diff
@@ -278,13 +278,18 @@ typedef enum {
...
-#define MXC_BASE_AESKEYS ((uint32_t)0x40205000UL)
-#define MXC_AESKEYS ((mxc_aeskeys_regs_t *)MXC_BASE_AESKEYS)
+/*                                                   System and User AES Keys */
+#define MXC_BASE_SYS_AESKEYS ((uint32_t)0x40205000UL)
+#define MXC_SYS_AESKEYS ((mxc_sys_aeskeys_regs_t *)MXC_BASE_SYS_AESKEYS)
+
+#define MXC_BASE_USR_AESKEYS ((uint32_t)0x40005000UL)
+#define MXC_USR_AESKEYS ((mxc_usr_aeskeys_regs_t *)MXC_BASE_USR_AESKEYS)
 
 // DEPRECATED(1-10-2023): Scheduled for removal.
```
Jake-Carter, line 288: Check if `@deprecated` works on macros as well?

Libraries/PeriphDrivers/Include/MAX32672/gpio.h
```diff
@@ -314,17 +314,6 @@ void MXC_GPIO_RegisterCallback(const mxc_gpio_cfg_t *cfg, mxc_gpio_callback_fn c
...
-/**
- * @brief      Set Voltage select for pins to VDDIO or VDDIOH
- *
- * @param      port   The GPIO port
- * @param[in]  vssel  VDDIO or VDDIOH to set the voltatge to
- * @param[in]  mask   Pins in the GPIO port that will be set to the voltage.
- * 
- * @return     #E_NO_ERROR if everything is successful. See \ref MXC_Error_Codes for the list of error codes.
- */
-int MXC_GPIO_SetVSSEL(mxc_gpio_regs_t *port, mxc_gpio_vssel_t vssel, uint32_t mask);
```
Jake-Carter, line 326: Added the "API Change" label since we removed this function.  Before merging we should add a note about this to the PR.  It's an unsupported feature of the micro, but still good practice to highlight this and why we did it for the next release notes.

Jake-Carter (approved):
Looks good.  I think we could improve the Doxygen output before merging

Libraries/CMSIS/Device/Maxim/MAX32672/Include/aeskeys_regs.h
sihyung-maxim: Ope, I didn't know @deprecated was actually used by Doxygen. I originally added the annotation into our deprecated register files, so the register workflows can filter out deprecated register files.

Libraries/CMSIS/Device/Maxim/MAX32672/Include/max32672.h
sihyung-maxim: It will throw an error if the c code sees the deprecated annotation. Will Doxygen know the macros are deprecated if you add a comment `// @deprecated` within the same line for each macro?

Jake-Carter: You should be able to add a Doxygen header for a preprocessor definition.  Though you may need the changes from #521 that enable Doxygen output for static members.  See [here](https://github.com/Jake-Carter/msdk/blob/0240928ae0077cb4651bbe52a4c526f63a71a2d0/Libraries/PeriphDrivers/Include/MAX32670/mxc_sys.h#L226) for an example where I did a doxygen header for a macro.

Ex:
```C
/*
 * @deprecated (description...)
 */
#define MXC_BASE_AESKEY MXC_BASE_AESKEYS
```

If you want to do it inline I think you have to do something like what we did for the enums.  Haven't tried this myself

```C
#define MXC_BASE_AESKEY MXC_BASE_AESKEYS /**< @deprecated (description) */
```

Jake-Carter:
@sihyung-maxim looks like this did the trick:  https://github.com/Analog-Devices-MSDK/msdk/pull/534/commits/252ba1cd5ea5e4978d63b3fe31b65835cd692203

We also were missing some `@ingroup` tags.  Do you think we could get the register generators updated with those changes in that commit?  Essentially we need to make sure every `@file` tag and register module has the correct top-level `@ingroup`.



Deprecated list works now:
image

Modules are correctly listed and show their associated header files:

image

Header files show their members:

image

Source code can be browsed:

image

Jake-Carter:
@sihyung-maxim also FYI I switched the drivers over to actually use the new header files:  https://github.com/Analog-Devices-MSDK/msdk/pull/534/commits/5cad636bf7514eb8b1d9afa123c94845be8cac3e

Jake-Carter:
> @sihyung-maxim also FYI I switched the drivers over to actually use the new header files: [5cad636](https://github.com/Analog-Devices-MSDK/msdk/commit/5cad636bf7514eb8b1d9afa123c94845be8cac3e)

This broke pretty much every other micro's builds.  (see https://github.com/Analog-Devices-MSDK/msdk/actions/runs/4641783585)

We'll need to add some target checking in `aes_revb.c`.  I suggest:

```C
#if TARGET_NUM == 32672
#include "sys_aeskeys_regs.h"
#include "usr_aeskeys_regs.h"
#else
#include "aeskeys_regs.h"
#endif
```

Libraries/CMSIS/Device/Maxim/MAX32672/Include/usr_aeskeys_regs.h
```diff
@@ -76,6 +77,7 @@ extern "C" {
 
 /**
  * @ingroup     usr_aeskeys
+ * @ingroup     aes
```
sihyung-maxim, line 80: Should the @ingroup annotation for all the aeskeys be `aes`?

Jake-Carter: Any `@ingroup` will be listed like this:

image

So I think a good criteria for what gets the "root" `aes` group is to match exactly what is included by the peripheral's header file.

Ex "aes.h":
image

Then the Doxygen browser will also give you clickable links that match 1:1 with its source code.

image

Jake-Carter (approved):
Approved to merge.  We should apply these Doxygen improvements to the register file generator

Jake-Carter (changes requested):
Sorry - scratch that.  I think we need to address https://github.com/Analog-Devices-MSDK/msdk/pull/534#issuecomment-1500694625

sihyung-maxim:
> > @sihyung-maxim also FYI I switched the drivers over to actually use the new header files: [5cad636](https://github.com/Analog-Devices-MSDK/msdk/commit/5cad636bf7514eb8b1d9afa123c94845be8cac3e)
> 
> This broke pretty much every other micro's builds. (see https://github.com/Analog-Devices-MSDK/msdk/actions/runs/4641813965/jobs/8215147365)
> 
> We'll need to add some target checking in `aes_revb.c`. I suggest:
> 
> ```c
> #if TARGET_NUM == 32672
> #include "sys_aeskeys_regs.h"
> #include "usr_aeskeys_regs.h"
> #else
> #include "aeskeys_regs.h"
> #endif
> ```

Looks like the aes_revb.c doesn't even need the AESKEYS registers, so I've removed all aeskey_regs.h inclusions.

Jake-Carter (approved):
Approved to merge when checks pass

Jake-Carter:
Manually running Build_Examples on this branch

https://github.com/Analog-Devices-MSDK/msdk/actions/runs/4877402221

sihyung-maxim:
> @sihyung-maxim Approved to merge when checks pass. The diff on the register file is against my doxygen changes. Are those easy to integrate into the file generator?

Yep, should be implemented after this PR is merged. 

Took **way** to long to figure out that the BTM-CI doesn't let you run shell scripts in the way the SVD workflow previously had used. I've updated all (minus ME13B and ME55) register files to include your new doxygen changes. Should be good now.

Jacob-Scheiffler (approved):

PR #547, opened 2023-04-11
GPIO RevA:  Protect Against GPIO Glitches on Configuring Alternate Function 

### LINKED ISSUE #533: MAX32672 - Alternate Function Configuration
The MXC_GPIO_RevA_SetAF() function needs to be updated to match the data from the User Guide.

image

I think that there is a typo on the procedure to select an alternate function. Step 3 it should reference table 6-5.
If this is the case the procedure should be the following:
1 - Set the transition acording to table 6-6
2 - Optional: Configure the electrical characteristics of the pin.
3 - Set the GPIO Mode acording to table 6-5

Is this procedure correct?

This steps require to modify three different registers. Is there a specific order to follow or the order doesn't matter?

BobbyCounts:
Always set EN0 first at entry and clear EN0 last when configuring alternate function modes.
closes #533


Jake-Carter (approved):
Looks good, thanks!

Re-merge with main to get the latest Build_Examples workflow as of #545.  That will resolve the build error check.

Jacob-Scheiffler (approved):

PR #549, opened 2023-04-13
MAX32675, MAX32680: AFE and HART Driver Updates from Bench and Simulation Testing for New Device Revision

jdk-maxim:


jdk-maxim:
Note: Seeing some strange behaviors with the check procedures where they report success but don't seem to actually be running.  
Guessing this may be related to this Pull Request coming from my Fork vs. a local branch.

For instance, on commit: [33b9867](https://github.com/Analog-Devices-MSDK/msdk/pull/549/commits/33b9867976421278b64f792fd7bb5f5e3351473b) reports all checks passed, but there was some cpplint issues identified, fixed in the following commit.

Details of the [Lint](https://github.com/Analog-Devices-MSDK/msdk/actions/runs/4693581391/jobs/8320736055?pr=549) show a warning of no files to lint.  Seems like a checkout issue, also not clear why the [WARN] did not bubble up to the top level as a yellow X or similar on the check summary instead of the green check-mark.

Similarly, they Build_Examples, Verify Register and SVD Files also do not seem to be working as expected.

jdk-maxim:
After converting from draft PR to "ready for review" now the BLE Examples Test check is running.  This was skipped for all previous runs on the draft PR.

EdwinFairchild:
This is running because 
`Enabled all tests to run because Libraries/PeriphDrivers/Source was  modified`
This will fail so I just cancelled it because the hardware is in the midst of being moved across the office and is all unplugged.

PR #552, opened 2023-04-14
MAX78000, MAX78002: Update Checkpoint for FaceID Database Generation

aniktash:
Updated the checkpoint used in database generation to match the one used to create weights.h 
Added seed initialization to make the embedding generation deterministic.


Jake-Carter (approved):
Seems reasonable to me, just need to find some time to test

vicloginov (approved):
Looks good

PR #568, opened 2023-04-26
MAX78000: Updating WearLeveling example to explicitly show flash wear leveling.

Jacob-Scheiffler:
This PR makes changes to the existing WearLeveling example to explicitly demonstrate the flash wear leveling ability of the LittleFS library. Additionally, it builds upon the previous example by adding a command line interface that allows users to read and write files interactively.

Note: Once this PR is approved and merged, this example will be ported to all other micros in a separate PR.

### DIFF
--- a/Examples/MAX78000/WearLeveling/include/main.h
+++ b/Examples/MAX78000/WearLeveling/include/main.h
@@ -1,0 +1,9 @@
+#ifndef MAIN_H_
+#define MAIN_H_
+
+/***** Definitions *****/
+#define LFS_START_PAGE 	48 	// First flash memory block used by LFS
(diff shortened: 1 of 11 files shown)


Jake-Carter (approved):

Examples/MAX78000/WearLeveling/include/main.h
```diff
@@ -0,0 +1,9 @@
+#ifndef MAIN_H_
```
sihyung-maxim, line 1: Ditto

sihyung-maxim (commented):
It looks like the copyrights are missing from the library files in the include/src folders. Are these files provided by a third-party or something we wrote?

sihyung-maxim (approved):

PR #570, opened 2023-04-27
FreeRTOS-Plus: Removing unused features with long path lengths.

Jacob-Scheiffler:
This PR removes coreMQTT, coreHTTP, FreeRTOS-Cellular-Interface, and coreMQTT-agent source code from the FreeRTOS-Plus library since they are unsupported by our micros and they contained files with path lengths greater than 160 characters. This will allow the SDK repo to be cloned into a directory that has a path length of 100 characters or less.


Jake-Carter (approved):
I think the only thing we officially support from FreeRTOS-Plus is FreeRTOS-Plus-CLI.   We could arguably remove everything else as well.

Jacob-Scheiffler:
I realized that all of the path lengths which exceed 160 characters were in test directories, so I have added back in the source code for the deleted feature and removed only the test folders.

sihyung-maxim (approved):

PR #571, opened 2023-04-28
feat(SDHC,MiscDrivers): Update AppSwitcher example to use new features with External Flash instead of SDHC

MertCanBoz1:
I changed the AppSwitcher (in "refdes" repository) example to work from the external flash instead of the SD Card. To be able to build the new code, some functions should be changed from here too.


Libraries/SDHC/ff13/Source/diskio.c
```diff
@@ -31,23 +33,44 @@ static unsigned int init_done = 0;
 
 /*local vaiables*/
 static uint8_t rtc_en;
-
+uint8_t flash = 0;
```
Jake-Carter, line 36: I am a little bit confused about using the global `flash` variable vs `#ifdef FLASH` in the preprocessor.

Consider the following function:

```C
DSTATUS disk_status (
    BYTE pdrv       /* Physical drive nmuber to identify the drive */
)
{
	if(!flash){
    	DSTATUS status = 0;

    	if (!MXC_SDHC_Card_Inserted()) {
			init_done = 0;
			status = STA_NOINIT | STA_NODISK;
    	}

    	return status;
	}
#ifdef FLASH
	else{
		    DSTATUS status = 0;


    if (mscmem_ID() == MX25_EXP_ID) {
    	status = 0;
    }
    else{
    	init_done = 0;
    	status = STA_NOINIT | STA_NODISK;
    }

    return status;
	}
#endif
	return 0;
}
```

Why is the global runtime variable necessary?  I don't see any calls to `disk_flash` in this driver code.  Given the compiler definition and the suggestion above we can reduce this function to:

```C
DSTATUS disk_status (
    BYTE pdrv       /* Physical drive nmuber to identify the drive */
)
{
    DSTATUS status = 0;
#ifndef FLASH
    init_done = MXC_SDHC_Card_Inserted();
#else
    init_done = mscmem_ID();
#endif
    if (!init_done) {
        status = STA_NOINIT | STA_NODISK;
    }
	return status;
}
```

Libraries/SDHC/ff13/Source/diskio.c
```diff
@@ -60,7 +83,36 @@ DSTATUS disk_initialize (
...
 )
 {
-    DSTATUS status;
+	if(!flash){
+    	DSTATUS status;
+	
+	    rtc_en = 0;
+#if (FF_FS_NORTC == 0)
+	    //Initialize RTC
+	    if (MXC_RTC->cn & MXC_F_RTC_CN_WE) {
```
Jake-Carter, line 92: Formatting issue on this block, make sure lines 92 - 101 are indented correctly

Libraries/SDHC/ff13/Source/diskio.c
```diff
@@ -101,17 +165,57 @@ DRESULT disk_read (
...
+	}
+#ifdef FLASH
+	else{
+		    DRESULT status;
+
+    int index = 0;
+
+    status = RES_OK;
+
+#if FF_MAX_SS == 4096
```
Jake-Carter, line 187: I am a little confused on the driver differences between `FF_MAX_SS = 4096` and `FF_MAX_SS = 512`.  

To improve the readability of the code I would suggest adding a comment explaining it, and you can also consolidate inside of the for loop:

```C
for(index = 0; index < count; index++){
#if FF_MAX_SS == 4096
    // ...
#endif
#if FF_MAX_SS == 512
    // ...
#endif
}
```

Libraries/SDHC/ff13/Source/diskio.c
```diff
@@ -146,33 +300,98 @@ DRESULT disk_ioctl (
...
+	
+    	return status;
+	}
+#ifdef FLASH
+	else{
+		    DRESULT status;
 
+
+    status = RES_OK;
+#if FF_MAX_SS == 4096
```
Jake-Carter, line 338: Same `FF_MAX_SS` considerations here - please add a comment explaining why it's necessary and consider consolidating the switch statements if possible

Libraries/SDHC/ff13/Source/diskio.h
```diff
@@ -14,9 +14,11 @@ extern "C" {
 #ifdef NATIVE_SDHC
 #include "sdhc_lib.h"
 #endif
-
+#ifdef FLASH
```
Jake-Carter, line 17: Arguably we should not modify the `diskio.h` header file at all.  It's provided by FatFS, and it is possible to handle all of the required logic within the implementation file.

Libraries/SDHC/ff13/Source/diskio.h
```diff
@@ -14,9 +14,11 @@ extern "C" {
...
+#include "mscmem.h"
+#endif
 #include "rtc.h"
-
+#include "integer.h"
```
Jake-Carter, line 21: `diskio.h` modification - Including `integer.h` should be done in the implementation file if necessary

Libraries/SDHC/ff13/Source/diskio.h
```diff
@@ -40,7 +42,7 @@ DRESULT disk_read (BYTE pdrv, BYTE* buff, DWORD sector, UINT count);
 DRESULT disk_write (BYTE pdrv, const BYTE* buff, DWORD sector, UINT count);
 DRESULT disk_ioctl (BYTE pdrv, BYTE cmd, void* buff);
 DWORD get_fattime(void);
-
+void disk_flash(uint32_t isFlash);
```
Jake-Carter, line 45: `diskio.h` modification - adding this function to the API should only be done as an absolute last resort.  Additionally, `uint32_t` is used which goes against the FatFS convention of using its built-in data types.

Libraries/SDHC/ff13/Source/ffconf.h
```diff
@@ -171,6 +174,8 @@
 
 #define FF_MIN_SS		512
 #define FF_MAX_SS		512
+//#define FF_MIN_SS		4096
```
Jake-Carter, line 177: This needs a comment explaining in what situation the sector size of 4096 would be used.

We should also consider whether we want to wrap this in some variable that we have more control over from the build system.

Ex:
```C
#ifdef SOMEVARIABLE
#define FF_MIN_SS		512
#define FF_MAX_SS		512
#else
#define FF_MIN_SS		4096
#define FF_MAX_SS		4096
#endif
```

Libraries/SDHC/ff13/Source/mscmem.c
```diff
@@ -0,0 +1,339 @@
...
+ * of trade secrets, proprietary technology, copyrights, patents,
+ * trademarks, maskwork rights, or any other form of intellectual
+ * property whatsoever. Maxim Integrated Products, Inc. retains all
+ * ownership rights.
+ *
+ ******************************************************************************/
+
+/**
+ * @file    mscmem.h
+ * @brief   Memory routines used by the USB Mass Storage Class example.
```
Jake-Carter, line 36: @sanjayjaroli can you take a look at the application of these files towards the SDHC library?  Are there any maintenance considerations?

@MertCanBoz1 since FatFS is not USB-specific I wonder if using the [Libraries/MiscDrivers/ExtMemory](https://github.com/Analog-Devices-MSDK/msdk/tree/main/Libraries/MiscDrivers/ExtMemory) is a better choice?

Libraries/SDHC/sdhc.mk
```diff
@@ -67,7 +67,7 @@ export PROJ_LDFLAGS
 export MXC_OPTIMIZE_CFLAGS
 export BOARD_DIR
 export USE_NATIVE_SDHC
-
+export FLASH
```
Jake-Carter, line 70: I suggest naming this variable `EXTERNAL_FLASH`

Jake-Carter (changes requested):

Libraries/SDHC/ff13/Source/diskio.c
MertCanBoz1: The global external_flash variable is necessary because the new AppSwitcher code will check whether the user uses an SD Card or the external flash (if the SD Card is not inserted, then continue with the external flash). I added a preprocessor because if an application other than AppSwitcher calls the diskio.c, it is not necessary to compile all mscmem functions. The disk_flash() is called from the main.c of the new AppSwitcher, which also exists as a PR in another repository (refdes).

Libraries/SDHC/ff13/Source/diskio.c
MertCanBoz1: I took the base code from the MAXREFDES104 project, and FF_MAX_SS = 4096 part was added to their code. I will remove all FF_MAX_SS = 4096 parts.

Libraries/SDHC/ff13/Source/diskio.h
MertCanBoz1: I am not sure how I can come around to not changing the diskio.h for this code snippet. I will look into it.

Libraries/SDHC/ff13/Source/mscmem.c
MertCanBoz1: It may be, but the external flash on the connectivity board of the MAXREFDES178 is different than the one used for EvKits. Some functions (specifically Ext_Flash_Erase.c ) need different handling than the ones at the mx25.c. The new AppSwitcher code calls mx25.c directly from the source file ( https://github.com/MaximIntegratedAI/refdes/pull/48 ).
Therefore, adding the mscmem.c to the [Libraries/MiscDrivers/ExtMemory](https://github.com/Analog-Devices-MSDK/msdk/tree/main/Libraries/MiscDrivers/ExtMemory) may cause some conflicts between them.

Libraries/SDHC/ff13/Source/diskio.c
Jake-Carter: Ok, thank you for clarifying.  This situation is a multi-drive system:

image
(http://elm-chan.org/fsw/ff/doc/appnote.html#port)

The [disk_initialize](http://elm-chan.org/fsw/ff/doc/dinit.html) function is already designed to handle this with the `pdrv` argument.  In the image above physical drive 2 is the SD card slot, and physical drive 0 is the external flash.

Following that scheme, the app switcher application code would first attempt to initialize drive 2 (SD).  If it fails, then you attempt to initialize drive 0 (Ext Flash).

The `diskio.c` file in the avr example in [ffsample.zip](https://github.com/Analog-Devices-MSDK/msdk/files/11392032/ffsample.zip) is a decent reference for a multi-drive implementation.

Notice how they are separating the driver layers based on a scheme they have defined for `pdrv`.

```C
/*-----------------------------------------------------------------------*/
/* Low level disk I/O module glue functions         (C)ChaN, 2016        */
/*-----------------------------------------------------------------------*/
/* If a working storage control module is available, it should be        */
/* attached to the FatFs via a glue function rather than modifying it.   */
/* This is an example of glue functions to attach various exsisting      */
/* storage control modules to the FatFs module with a defined API.       */
/*-----------------------------------------------------------------------*/

#include "ff.h"			/* Obtains integer types for FatFs */
#include "diskio.h"		/* FatFs lower layer API */
#ifdef DRV_CFC
#include "cfc_avr.h"	/* Header file of existing CF control module */
#endif
#ifdef DRV_MMC
#include "mmc_avr.h"	/* Header file of existing SD control module */
#endif

// ...

/*-----------------------------------------------------------------------*/
/* Inidialize a Drive                                                    */
/*-----------------------------------------------------------------------*/

DSTATUS disk_initialize (
	BYTE pdrv				/* Physical drive nmuber to identify the drive */
)
{
	switch (pdrv) {
#ifdef DRV_CFC
	case DRV_CFC :
		return cf_disk_initialize();
#endif
#ifdef DRV_MMC
	case DRV_MMC :
		return mmc_disk_initialize();
#endif
	}
	return STA_NOINIT;
}
```

Libraries/SDHC/ff13/Source/diskio.h
Jake-Carter: See https://github.com/Analog-Devices-MSDK/msdk/pull/571/files#r1184416416

Libraries/SDHC/ff13/Source/mscmem.c
Jake-Carter: I think that's what I mean - we should use mx25.c directly.  Ext_Flash.h provides a struct to abstract the physical communication layer which should be enough to address connectivity issues I think.  

https://github.com/Analog-Devices-MSDK/msdk/blob/9bf68cf780920cc3d9654418c5722543f2e69c89/Libraries/MiscDrivers/ExtMemory/Ext_Flash.h#L143

Maybe I missed something - Does the cube camera present itself as a mass storage device for USB?  Is that what the mscmem files are used for?

Jake-Carter (approved):
Looks good, thanks for implementing the changes.

I have one last minor suggestion below, but otherwise I think it's ready.  Nice work!

sihyung-maxim:
Not sure why the workflows aren't running...

sihyung-maxim (approved):

sihyung-maxim (approved):

PR #574, opened 2023-05-01
added BLE5_ctr CI test

yc-adi:
BLE5_ctr projects are used in RF-PHY-closed for PER and timing tests. But msdk repo changes could corrupt the BLE5_ctr. So this CI test is added to verify the changes related to BLE5_ctr.

### DIFF
--- a/.github/workflows/ble5_ctr_test.yml
+++ b/.github/workflows/ble5_ctr_test.yml
@@ -27,6 +27,13 @@ jobs:
           submodules: false
           repository: Analog-Devices-MSDK/msdk
           ref: main
+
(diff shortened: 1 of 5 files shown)


yc-adi:
The test result showed that the current msdk MAX32655/BLE5_ctr won't work.

EdwinFairchild:
@yc-adi  I wonder if this is because of the fact they have the new  ME17B chips.

EdwinFairchild:
Yeah, I think there is a lot of debugging to do with ME17B Phy code this is why also some examples do not work correctly. Still waiting on those boards with new chip.

PR #580, opened 2023-05-08
RTC: Deprecating MXC_RTC_GetSubSecond and MXC_RTC_GetSecond

Jacob-Scheiffler:
This PR resolves JIRA ticket MSDK-1106 by deprecating the RTC functions GetSubSecond and GetSecond, and replacing them with GetSubSeconds and GetSeconds.

GetSubSecond/GetSecond were implemented incorrectly because the return value was used both to indicate whether an error had occurred and to return the count values of either the SEC or SSEC register. This was especially problematic for the SEC register since it is a 32-bit unsigned integer. Returning an error from GetSecond would aliases to a large SEC count, thus it was impossible to determine whether an error had occurred or the count value was simply a large value. (Note: this was not a bug for GetSubSecond since it is only a 12-bit unsigned value, but for consistency it has been deprecated too.)

The new functions solve this issue by taking a pointer to a uint32_t variable as a parameter and storing the current count of either the SSEC or SEC register to that address and the return value is now only used to indicate success/failure.

Note: All RTC and RTC_Backup exampless were updated to use the new functions.


Libraries/PeriphDrivers/Include/MAX32670/rtc.h
```diff
@@ -177,19 +177,35 @@ int MXC_RTC_GetFlags(void);
...
  */
-int MXC_RTC_GetSecond(void);
+int MXC_RTC_GetSubSeconds(uint32_t *ssec);
 
 /**
- * @brief     Get the time using nuclear fusion. Or atomically. Something like that.
+ * @brief     Get Second or E_BUSY, see /ref MXC_ERROR_CODES
+ * @retval    returns second value
+ */
+int MXC_RTC_GetSecond(void) __attribute__((deprecated("Use MXC_RTC_GetSeconds() instead.")));
```
Jake-Carter, line 197: Looks good, we just need to wrap the attributes in a check for GCC so we don't break IAR/Keil

```C
/**
 * @brief     Get Second
 * @retval    returns second value or E_BUSY, see /ref MXC_ERROR_CODES
 */
#ifdef __GNUC__
__attribute__((deprecated("Use MXC_RTC_GetSeconds() instead.")))
#endif
int MXC_RTC_GetSecond(void);
```

Jake-Carter (changes requested):

Jake-Carter (approved):

Jake-Carter (approved):

sihyung-maxim (approved):

PR #586, opened 2023-05-12
Add die type to part number conversion table to User Guide

jlj-ee:
The VSCode-Maxim repo used to have a page in the wiki that showed the die type to part number conversion, which was useful for navigating the peripheral driver code. 

When the wiki was moved and its content merged with the User Guide, that section appears to have not been migrated. This PR adds the same conversion table from the old wiki to [the 'Peripheral Driver API > Organization' section of the User Guide](https://analog-devices-msdk.github.io/msdk/USERGUIDE/#organization). 

To build properly, `mkdocs.yml` was also modified slightly to enable the markdown table extension. 

The `Documentation\build.py` script was run locally and a screenshot of the relevant section from the generated `docs\USERGUIDE\index.html` is shown below:

userguide_table

### DIFF
--- a/USERGUIDE.md
+++ b/USERGUIDE.md
@@ -1706,8 +1706,8 @@ The _**implementation**_ files are further organized based on _**die type**_ and
       | MAX32690 | ME18 |
       | MAX78000 | AI85 |
       | MAX78002 | AI87 |
-      
-- The **_hardware** revision_ files follow the **`_revX`** naming convention.  
+
+- The **_hardware revision_** files follow the **`_revX`** naming convention.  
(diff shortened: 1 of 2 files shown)


jlj-ee:
I actually see a misplaced formatting flag for "hardware revision" in the screenshot above that is present in the current hosted version of the User Guide. Since it is part of the section in question, I will update this PR with a commit to fix that as well.

Jacob-Scheiffler (approved):

Jake-Carter:
Online copy updated as of https://github.com/Analog-Devices-MSDK/msdk/actions/runs/5007625758/jobs/8974474872

PR #587, opened 2023-05-16
MiscDrivers: Update all boards to use PB, LED, and stdio source files in MiscDrivers.

Jacob-Scheiffler:
This PR removes all Include and Source directories in the Boards/<Part Number> and updates all board.mk files to use MiscDrivers. Most notably this affects the LED, PB, and stdio paths/files, but some of the boards used local copies of other drivers supported in MiscDrivers and these were updated as well. As well, some minor tweaks had to be made in the LED, PB , and stdio drivers to make them compatible with all parts.


Libraries/MiscDrivers/PushButton/pb.c
```diff
@@ -74,18 +75,26 @@ int PB_RegisterCallback(unsigned int pb, pb_callback callback)
...
-void GPIO1_IRQHandler(void)
-{
-    MXC_GPIO_Handler(MXC_GPIO_GET_IDX(MXC_GPIO1));
-}
-void GPIO2_IRQHandler(void)
+/******************************************************************************/
+int PB_RegisterCallbackRiseFall(unsigned int pb, pb_callback callback)
 {
-    MXC_GPIO_Handler(MXC_GPIO_GET_IDX(MXC_GPIO2));
+    MXC_ASSERT(pb < num_pbs);
```
sihyung-maxim, line 81: Instead of asserting, you can also return an error code if incorrect `pb` parameter, so this function can follow the doxygen comments and not have one return code.

sihyung-maxim (approved):

sihyung-maxim:
@EdwinFairchild @EricB-ADI @yc-adi 

Not sure if any of the BLE codes use the PB library, but the BLE team should also review these changes since GPIO interrupt handling was updated.

EdwinFairchild:
GPIO interrupt checks out

EdwinFairchild (approved):

Jake-Carter (approved):

PR #588, opened 2023-05-17
MiscDrivers: Adds volatile qualifier to variable used in ext flash

EdwinFairchild:
Prevents the timeout variable from being optimized out.
Not sure why now and not before, but I am seeing consistent failed write page operations during OTA because of
"busy" state. This fixes the issue.


Jacob-Scheiffler (approved):
External flash erases take a long a** time; we should probably double check that the delays are long enough too, especially for the MX25.

sihyung-maxim (approved):

sihyung-maxim:
> External flash erases take a long a** time; we should probably double check that the timeouts are long enough too, especially for the MX25.

Is this something we wanna verify before merging this PR?

EdwinFairchild:
Yeah ill verify later this afternoon

Jake-Carter:
@EdwinFairchild good to merge?

EdwinFairchild:
yeah tested both and erase/write operations are fine.

PR #590, opened 2023-05-18
lwIP: Add OS mode configuration

ttmut:
## Overview

This PR adds support to use lwIP in OS mode. Existing lwIP examples run as bare metal applications (NO_SYS=1). A few modifications are necessary to make lwIP run on an OS.

- Default configuration in _lwipopts.h_ disables OS mode by setting NO_SYS to 1. It can be overridden using compiler flags.
       `PROJ_CFLAGS += -DNO_SYS=0`
- A few settings regarding thread and mailbox sizes are added into _lwipopts.h_. They are ignored by lwIP in NO_SYS mode.
- lwIP is now initialized using `tcpip_init` in OS mode as suggested by lwIP docs.
- Use system provided _errno.h_ header and `timeval` struct.

Note that besides above changes, OS abstraction layer functions (see _<lwip/sys.h>_ for prototypes) must be implemented by the user. I will provide an example that runs on freeRTOS.

## Linked Tickets

MSDK-1143
MSDK-1144

## Key Changes

I had to include _<sys/errno.h>_ instead of _<errno.h>_ because the build system kept using _<lwip/errno.h>_ instead of the system provided one.

## Tests

Tested existing lwIP_Ping and lwIP_TCP examples to make sure existing projects are not broken by the changes. lwIP_Ping example is then reimplemented to work on top of freeRTOS using lwIP OS mode.


ozersa:
@Jake-Carter  how about excluding lwip stack linter format check?
https://github.com/Analog-Devices-MSDK/msdk/blob/main/.github/workflows/linter.yml#L62

ozersa (approved):
Approved

Jake-Carter (approved):

sihyung-maxim (approved):

sihyung-maxim (approved):

PR #597, opened 2023-05-26
feat(PeriphDrivers): Add GPIO3 support to MAX32655, MAX32680, MAX78000, and MAX78002

Jacob-Scheiffler:
The Always-On-Domain GPIO port (GPIO3) was only partially supported on the MAX32665, MAX32680, MAX78000, and MAX78002. This PR adds complete support for the AOD GPIO ports so they may be used transparently with our standard GPIO API.

NOTE: GPIOWAKE_IRQn was added to MXC_GPIO_GET_IRQ for each of these chips in [PR #587](https://github.com/Analog-Devices-MSDK/msdk/pull/587).


Jacob-Scheiffler:
Looks like solving the Bootloader example build errors will not be a trivial fix, so I added the WIP label while we work on the solution.

EdwinFairchild (approved):

sihyung-maxim (approved):

PR #600, opened 2023-05-31
feat(Examples,CMSIS,PeriphDrivers): Update HART UART with new transmit features and fixes for MAX32675 and MAX32680

jdk-maxim:
Add additional requested features to HART UART.

- Interrupt driver HART transmit.  API requires notification of transmit complete, using timer to count bit-times, and arm a callback interrupt to verify transmit is complete.  Originally implemented to use MXC_DelayAysnc, but discovered some issues with this, with small delays not getting called back correctly, and no nesting of delays allow.  Therefore, using a dedicated system timer, to minimize resource usage, split it into two 16 bit timers, one for AFE SPI timeouts, and the other for HART_UART.  Implemented support for this in afe_timer.c as current timer API doesn't fully support split timers.
- Service Access Points (SAP), as defined in HART Token Passing Data Link Layer (TPDLL) specification, these functions and callbacks provided a standard break between customer stacks and this physical layer HAL shim.
- Several other fixes and improvements.
- HART_UART examples updated to show and test new features.

Also a couple of incidental fixes for TX_OB, ME16 Timer Warnings, and Change to GPIO IRQ handlers.


jdk-maxim:
Added Jake and Sihyung regarding the Action failure.

jdk-maxim:
@ozersa  This includes changes based on your comments from previous [Pull Request](url).  I also added some of the downstream changes from the uPython port you [previously mentioned](https://github.com/Maxim-Central-Software-Services/abb-max32675/commit/ba418b94b5c8c8dfdd98dc9af8fdaa568f16fd5e) as well.  I elected to not include the setup with specific frequency, and clock divider params as, these were passed to MXC_UART_RevB_SetFrequency_Custom which doesn't exist in the current SDK.  I also did not bring in the clkdiv param to enable_hart_clock.  This is complicated by the MAX32680 having to use the Pulse Train for clock gen and doesn't map directly to clock divider.  So I left it alone for now.

sihyung-maxim (approved):

sihyung-maxim:
Fixed the Verify_Register workflow. Nothing wrong on your part, the machine that hosts the workflow's runner never reran the SVD scripts when checking with your register files. Weird bug, but I'll fix it in another PR.

I'll merge in thus PR once the workflows all finish.

PR #604, opened 2023-06-01
Removed wrong instruction

Suraj-Ajjampur:
Removed extra instruction from the UART example file.
"Connect the jumper (JP7) to UART0."


EricB-ADI (commented):
Why is this wrong? The schematic has JP7 as a pullup to select the UART through a MUX. If you disconnect the jumper it will be floating and may or may not work.

Suraj-Ajjampur:
Clarified as a documentation issue in the Datasheet. The README.md file gives the right instructions. Shall raise a ticket for the same.

EricB-ADI:
Im going to go ahead and close this PR

PR #615, opened 2023-06-10
feat(BLE,Examples): Add Cordio API to get Bluetooth MAC address, trace output of MAC address and advertising name for MAX32655, MAX32665, MAX32690

EdwinFairchild:
This fixes random MAC address issue on ME17 and I think it is helpful to print the MAC address on start up. Helped me find a bug where all ME17 had different MAC addresses due to some junk value always being put in there.


EdwinFairchild:
@yc-adi @EricB-ADI 
❤️

EricB-ADI:
Is the or unique enough?

EdwinFairchild:
those are actually the x,y coordinates on the wafer

EricB-ADI:
> those are actually the x,y coordinates on the wafer

Hmm did not know that

yc-adi (approved):

PR #622, opened 2023-06-14
docs(Examples): Delete outdated board info in MAX78000 Hello_World README.md

Suraj-Ajjampur:
Documented in [MSDK-1203](https://jira.maxim-ic.com/browse/MSDK-1203?orderby=created+DESC%2C+priority+DESC%2C+updated+DESC)


Jake-Carter (approved):

Jacob-Scheiffler (changes requested):
The BOARD setting needing to be changed should still be documented in the README(s). Eclipse users still have to make the change in project.mk, so I think something like, "Select "FTHR_RevA" for _BOARD_ in "project.mk" for Eclipse users or in "settings.json" for VSCode users." would be better.

Jake-Carter:
@Suraj-Ajjampur We already have the relevant info in the readme (see above comment), so commit https://github.com/Analog-Devices-MSDK/msdk/pull/622/commits/3f38a139646ebdcb59162de89a832049ba71725b is unnecessary.  Just remove the two lines from the hardware setup instructions.

PR #631, opened 2023-06-16
fix(Examples): MSDK-1192: Fix Leading Spaces issue in the Flash CLI example

Vignesh-VVadivel:
1. Fixed the leading spaces issue in the Flash CLI (1st issue in the ticket).  
2. Removed magic numbers and included relevant ASCII equivalent with Macros.  
3. Included few more check to avoid unwanted echo statements (2nd issue in the ticket).  

Corresponding Jira ticket is [here](https://jira.maxim-ic.com/browse/MSDK-1192)  

Note: - Line 236 (index != 0) in the updated code would resolve the second issue in the ticket.

### DIFF
--- a/Examples/MAX78000/Flash_CLI/main.c
+++ b/Examples/MAX78000/Flash_CLI/main.c
@@ -154,7 +154,7 @@ void checkLeadingSpaces(char *buffer, unsigned int *index)
             break;
     }
     if (leadingZerosCount > 0) {
-        memcpy(buffer, buffer + leadingZerosCount, (*index) - leadingZerosCount);
+        memmove(buffer, buffer + leadingZerosCount, (*index) - leadingZerosCount);
         *index = *index - leadingZerosCount;
         buffer[*index] = NULL_TERMINATION;
         ++(*index);
--- a/Examples/MAX78000/Flash_CLI/main.c
+++ b/Examples/MAX78000/Flash_CLI/main.c
@@ -73,9 +73,16 @@ TaskHandle_t cmd_task_id;
 #define STRING_(x) #x
 
 /* ASCII macros */
+#define END_OF_TEXT 0x03
 #define BACK_SPACE 0x08
 #define SPACE_BAR 0x20
 #define NULL_TERMINATION 0x00
+#define ARROW_KEY_CODE_1 0x1B
+#define ARROW_KEY_CODE_2 0x5B
+#define ARROW_KEY_CODE_LEFT 0x44
+#define ARROW_KEY_CODE_RIGHT 0x43
+#define ARROW_KEY_CODE_UP 0x41
+#define ARROW_KEY_CODE_DOWN 0x42
 
 /* Console ISR selection */
 #if (CONSOLE_UART == 0)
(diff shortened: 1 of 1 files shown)


Examples/MAX78000/Flash_CLI/main.c
```diff
@@ -184,11 +215,11 @@ void vCmdLineTask(void *pvParameters)
         if (async_read_req.rxCnt > 0) {
             /* Process character */
             do {
-                if (tmp == 0x08) {
+                if (tmp == BACK_SPACE) {
                     /* Backspace */
                     if (index > 0) {
                         index--;
-                        printf("\x08 \x08");
+                        printf("%c %c", BACK_SPACE, BACK_SPACE);
```
Jake-Carter, line 222: I would also suggest deleting the last character in the buffer when backspace is received.  In this case we can replace the value with the null terminator to "delete" it.

Jake-Carter (changes requested):
Looks good.  Minor change request below.  Do you also want to filter out arrow keys in this PR?

Examples/MAX78000/Flash_CLI/main.c
Vignesh-VVadivel: Ya sure. Null terminator would make more sense. Will do that change.

Vignesh-VVadivel:
> Looks good. Minor change request below. Do you also want to filter out arrow keys in this PR?

Yes. I want to filter out the arrow keys in this PR.
I did the coding part, to filter out the arrow keys. I am testing the code thoroughly and will push the changes ASAP.

Examples/MAX78000/Flash_CLI/main.c
```diff
@@ -200,26 +260,42 @@ void vCmdLineTask(void *pvParameters)
...
                     buffer[index++] = tmp;
-                    fflush(stdout);
+                    bool arrowKey = false;
+                    if (index >= 3) {
+                        arrowKey = checkArrowKeys(buffer + index - 3);
+                    }
+                    if (!arrowKey) {
+                        /* Echo out if not an arrow key */
+                        if ((tmp != ARROW_KEY_CODE_1) && (tmp != ARROW_KEY_CODE_2))
+                            putchar(tmp);
```
Jake-Carter, line 292: I think we can avoid the second check on line 291 if we reorder things slightly.

```C
bool arrowKey = false;
if (index >= 3) {
    arrowKey = checkArrowKeys(buffer + index - 3);
    if (arrowKey) { //Remove from buffer }
}
if (!arrowKey) {
    putchar(tmp);
    fflush(stdout);
}

**Review by Jake-Carter — CHANGES_REQUESTED** [review]
Looks good!  One final suggestion below

File: Examples/MAX78000/Flash_CLI/main.c (hunk 2 above)
Comment (Vignesh-VVadivel) [reply Examples/MAX78000/Flash_CLI/main.c]: That check is needed because over UART we receive only one byte of data through interrupts. For arrow inputs, we receive three bytes of data, which eventually triggers the interrupt three times. This would lead to few more undesirable echo statements.

And so, I included it, as a part of sanity check.  

NOTE:- This example code does not support **[** input, so custom commands shall not use this symbol.

**Review by Jake-Carter — APPROVED** [review]

**Review by sihyung-maxim — APPROVED** [review]

File: Examples/MAX78000/Flash_CLI/main.c [hunk 3]
```diff
@@ -122,6 +135,53 @@ void vCmdLineTask_cb(mxc_uart_req_t *req, int error)
...
+ */
+void checkLeadingSpaces(char *buffer, unsigned int *index)
+{
+    unsigned int leadingZerosCount = 0;
+    for (leadingZerosCount = 0; leadingZerosCount < (*index); leadingZerosCount++) {
+        if (buffer[leadingZerosCount] != SPACE_BAR)
+            break;
+    }
+    if (leadingZerosCount > 0) {
+        memcpy(buffer, buffer + leadingZerosCount, (*index) - leadingZerosCount);
```
lorne-maxim, line 157: memcpy is not guaranteed to handle overlapping src and dest buffers.  This probably works with our default standard libraries, but other versions may not.  memmove is a better choice.

lorne-maxim (changes requested):

Examples/MAX78000/Flash_CLI/main.c
Vignesh-VVadivel: Sure. Updated the code with memmove() to avoid overlapping buffers and tested it. 
Working fine now and I pushed it.

lorne-maxim (approved):

PR #635, opened 2023-06-22
fix(Examples): MSDK-1208 Fix I2C Pin number for Audio Codec

Vignesh-VVadivel:
As per the User guide, **I2C_SDA** pin of Audio Codec is connected to Port_Pin 0_17, **I2C_SCL** pin of Audio Codec is connected to Port_Pin 0_16. As per the description, **I2C1** is connected to P0_16 and P0_17.

And so, to configure the I2C pin for Audio Codec we should **use I2C1** rather than I2C0.
i2c_user_guide
i2c_description

The corresponding Jira ticket is [here](https://jira.maxim-ic.com/browse/MSDK-1208)


Jake-Carter (approved):
Good catch!

Jacob-Scheiffler (approved):

sihyung-maxim (approved):

PR #643, opened 2023-06-28
fix(Examples)!: Require DMA Channel as Input Argument for ADC DMA Conversions, Fix ADC DMA Transfer Bug for Channels > 0

Vignesh-VVadivel:
Fixed the DMA transfer over multiple channels for ADC example in the MAX32672 board, and the corresponding JIRA ticket is [here](https://jira.maxim-ic.com/projects/MSDK/issues/MSDK-1218?filter=allopenissues)

This PR contains the fix for all example codes that uses ADC Rev B, which includes MAX32662, MAX32672, MAX32690, MAX78002.

### DIFF
--- a/Libraries/PeriphDrivers/Include/MAX32662/adc.h
+++ b/Libraries/PeriphDrivers/Include/MAX32662/adc.h
@@ -202,7 +202,7 @@ typedef struct {
     mxc_adc_avg_t avg_number; ///< no of samples to average
     mxc_adc_div_lpmode_t lpmode_divder; ///< Divide by 2 control in lpmode
     uint8_t num_slots; ///< num of slots in the sequence
-    int8_t dma_channel; ///< User should call MXC_DMA_AcquireChannel() to assign DMA Channel number for this field
+    int8_t dma_channel; ///< User should call MXC_DMA_AcquireChannel() to assign DMA Channel number
 } mxc_adc_conversion_req_t;
 
 /**
--- a/Libraries/PeriphDrivers/Include/MAX32662/adc.h
+++ b/Libraries/PeriphDrivers/Include/MAX32662/adc.h
@@ -202,7 +202,10 @@ typedef struct {
     mxc_adc_avg_t avg_number; ///< no of samples to average
     mxc_adc_div_lpmode_t lpmode_divder; ///< Divide by 2 control in lpmode
     uint8_t num_slots; ///< num of slots in the sequence
-    int8_t dma_channel; ///< User should call MXC_DMA_AcquireChannel() to assign DMA Channel number
+    int8_t dma_channel; /*!<The channel to use for DMA-enabled transactions.
+                            Users should call \ref MXC_DMA_AcquireChannel
+                            to assign the DMA channel number.
+                            For non-DMA transactions, this field is ignored.*/
 } mxc_adc_conversion_req_t;
 
 /**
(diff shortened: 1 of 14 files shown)


Libraries/PeriphDrivers/Include/MAX32672/adc.h
```diff
@@ -246,6 +246,7 @@ typedef struct {
     mxc_adc_avg_t avg_number; ///< no of samples to average
     mxc_adc_div_lpmode_t lpmode_divder; ///< Divide by 2 control in lpmode
     uint8_t num_slots; ///< num of slots in the sequence
+    int8_t dma_channel; ///< DMA channel number for DMA transfer
```
Jacob-Scheiffler, line 249: We should consider updating this comment to say that users should call MXC_DMA_AcquireChannel to assign the value for this field.

Libraries/PeriphDrivers/Source/ADC/adc_revb.c
```diff
@@ -214,7 +214,7 @@ int MXC_ADC_RevB_StartConversionDMA(mxc_adc_revb_regs_t *adc, mxc_adc_conversion
 
     num_bytes = (req->num_slots + 1) * 4; //Support 8 slots (32 bytes) only. (TODO)
 
-    channel = MXC_DMA_AcquireChannel();
+    channel = req->dma_channel;
```
Jacob-Scheiffler, line 217: This change will break builds for the other chips which use ADC Rev. B. If we proceed with this implementation these changes will need to be applied to the ME12 and ME18 as well.

Jacob-Scheiffler (changes requested):

Libraries/PeriphDrivers/Include/MAX32672/adc.h
Vignesh-VVadivel: Updated that comment in the code.

Libraries/PeriphDrivers/Source/ADC/adc_revb.c
Vignesh-VVadivel: Yup sure,
I have updated these changes for both ME12 and ME18. 
I have also implemented these changes for MAX78002, since it uses ADC Rev B.

Jacob-Scheiffler (approved):

Libraries/PeriphDrivers/Include/MAX32662/adc.h
```diff
@@ -202,6 +202,7 @@ typedef struct {
     mxc_adc_avg_t avg_number; ///< no of samples to average
     mxc_adc_div_lpmode_t lpmode_divder; ///< Divide by 2 control in lpmode
     uint8_t num_slots; ///< num of slots in the sequence
+    int8_t dma_channel; ///< User should call MXC_DMA_AcquireChannel() to assign DMA Channel number
```
Jake-Carter, line 205: I would suggest changing this comment to indicate that the field is ignored for non-DMA transactions.
```C
///<The channel to use for DMA-enabled transactions.  Users should call \ref MXC_DMA_AcquireChannel to assign the DMA channel number.  For non-DMA transactions, this field is ignored.
```

It looks like we also need to improve our Doxygen headers for the TransactionDMA function in general.  Can you improve the adc.h files to fix the following issues?
1. The Doxygen for `mxc_adc_req_t` is broken (no link/documentation generated)
2. Update the `MXC_ADC_StartConversionDMA` function brief to inform the user that they must acquire a DMA channel and pass it in via the "req" input struct
3. Update the `MXC_ADC_StartConversionDMA` function brief to inform the user that they must enable the correct DMA IRQ call the `MXC_DMA_Handler` function from the ISR.

Jake-Carter (changes requested):

Jake-Carter (changes requested):
Thanks @Vignesh-VVadivel,

Final task for you on this:  `mxc_adc_req_t` and `mxc_adc_slot_req_t` are also missing Doxygen `@brief` sections for:
* MAX32662 `adc.h`
* MAX32672 `adc.h`
* MAX32690 `adc.h`
* MAX78002 `adc.h`

We should add them so that these structs also show up in the structs section:
image

Jacob-Scheiffler (approved):

Jake-Carter (approved):
Looks good, thanks

PR #647, opened 2023-07-03
fix(workflow): Make pull_request_template.md visible and update trigger-types for auto-labeler workflow

sihyung-maxim:
The "Documentation" folder is not a valid directory to place the pull_request_template.md (Docs is valid though). I moved the template into `.github/`, and hopefully, it does not override the README.md in the repo's root.

I also added more trigger types for the auto-labeler to run and removed the concurrency because it's a lightweight action that runs from Github's Ubuntu servers. (And in case the workflow gets cancelled because we have a lot of actions running for a PR now).


EdwinFairchild (approved):

.github/workflows/labeler.yml
```diff
@@ -1,13 +1,9 @@
...
 
-concurrency: 
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
-
 # Controls when the workflow will run
 on:
   pull_request_target:
-    types: [opened, edited, reopened]
+    types: [opened, edited, reopened, synchronize, converted_to_draft, readu_for_review, assigned, review_requested]
```
Jacob-Scheiffler, line 6: ready_for_review?

sihyung-maxim: oops, good catch

Jacob-Scheiffler (approved):

PR #648, opened 2023-07-03
fix(BLE,Examples): Log packet count on BLE dats speed test

EdwinFairchild:
Adds some visual feedback that packets are indeed being received on the sever side. Someone emailed me about throughput and was wondering if packets are indeed being received on the other side in this speed test.


EricB-ADI (approved):

EdwinFairchild:
@yc-adi

yc-adi (approved):

PR #662, opened 2023-07-11
feat(Examples,PeriphDrivers)!: Add SPI v2 Target Transaction functions

sihyung-maxim:
### Description

- Added Target Transaction feature to SPI v2 drivers.
- Updated SPI_MasterSlave example for MAX78002 to use SPI v2 drivers. (Showcases SPI v2 working with the previous SPI API).
- Created SPI_ControllerTarget example for MAX78002 to demo the new Controller/Target features on SPI v2.
- Fixed DMA burst size to correspond with TX/RX Threshold as per design requirements. (Threshold + 1 = DMA burst size).
- Blocking SPI controller (L. Master) transaction function is no longer interrupt-driven.
- Clarification: The controller (master) callback function is called within the interrupt handler at the end of the transaction (unlike RevA1).
- Fixed AI87 NewHaven touchscreen support.
- Default SPI build to SPI v1.
- Reverted `mxc_spi_req_t` struct to old names.
- Removed SW_DRV TS Control Scheme. Too much overhead and complexity.

Breaking change because there are functional differences with the interrupt handling between the previous SPI API and SPI v2 even if the function prototypes remain the same.

### Estimated Timing Averages from function call to start of transaction:
SYSCLK = IPO

#### SPI v2
Before Excess SPI Instance Check/Asserts Removal:
- ~30us - non-DMA Controller Transaction
- ~19us - DMA Controller Transaction

After Excess SPI Instance Check/Asserts Removal:
- ~27us - non-DMA Controller Transaction
- ~16us - DMA Controller Transaction

SPI v1
- ~57us - non-DMA Master Transaction
- ~135us - DMA Master Transaction

Final Overhead Improvement:
- ~52% non-DMA transaction
- ~88% DMA transaction

### Library Size:
**Before Excess SPI Instance Check/Asserts Removal**
image

**After Excess SPI Instance Check/Asserts Removal**
image

Tests:
- [x] Running Controller/Target transaction functions on two different SPI instances on the same board.
- [x] Running Controller/Target transaction functions on two different boards (one as a controller and one as a target).

- [ ] (Optional) Link any related Github issues [using a keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)


sihyung-maxim:
For consistency, should I rename SPI_MasterSlave to SPI_ControllerTarget?

sihyung-maxim:
Also, I just copy-pasted the porting guide/README from the SPI example to the SPI_MasterSlave. I am not sure if this is good enough, so please review it and let me know if it's enough for a user to follow.

sihyung-maxim:
> For consistency, should I rename the SPI_MasterSlave example to SPI_ControllerTarget?

I renamed it.

sihyung-maxim:
> > For consistency, should I rename the SPI_MasterSlave example to SPI_ControllerTarget?
> 
> I renamed it.

Another idea. Should I leave the SPI_MasterSlave example using the SPI v2 Master/Slave functions to show the functional changes between the previous SPI API and SPI v2?

sihyung-maxim:
Ready for review!

Jake-Carter:
@sihyung-maxim cherry-pick commit [31f54fe1c0a7af7db9076827183cec8e28ddc541](https://github.com/Analog-Devices-MSDK/msdk/commit/31f54fe1c0a7af7db9076827183cec8e28ddc541) to retrieve the Builld_Examples/Regression_Test_Make_3_81 workflow

Libraries/PeriphDrivers/Include/MAX78002/spi.h
```diff
@@ -1039,6 +1028,48 @@ int MXC_SPI_SlaveTransactionAsync(mxc_spi_req_t *req);
...
+ * 
+ * @param   spi         Pointer to SPI instance's registers.
+ * @param   tx_buffer   Pointer to transmit buffer (in terms of bytes).
+ * @param   tx_fr_len   Number of frames to transmit from transmit buffer.
+ * @param   rx_buffer   Pointer to transmit buffer (in terms of bytes).
+ * @param   rx_fr_len   Number of frames to store in recieve buffer.
+ *
+ * @return  See \ref MXC_Error_Codes for the list of error return codes.
+ */
+int MXC_SPI_TargetTransaction(mxc_spi_regs_t *spi, uint8_t *tx_buffer, uint32_t tx_fr_len,
```
Jake-Carter, line 1042: I don't think it makes sense to make this non-blocking.  `SPI_TargetTransaction` should be a replacement for `SPI_MasterTransaction`, which is blocking.

Jake-Carter, line 1042: We're changing the arguments from a request struct to a list of parameters.
1. This voids `TargetTransaction` as a drop-in replacement for `MasterTransaction`.
2. We will have problems in the future if/when we need to modify these arguments.  The previous struct gives us that flexibility.

Libraries/PeriphDrivers/Include/MAX78002/spi.h
```diff
@@ -1039,6 +1028,48 @@ int MXC_SPI_SlaveTransactionAsync(mxc_spi_req_t *req);
...
+ * 
+ * @param   spi         Pointer to SPI instance's registers.
+ * @param   tx_buffer   Pointer to transmit buffer (in terms of bytes).
+ * @param   tx_fr_len   Number of frames to transmit from transmit buffer.
+ * @param   rx_buffer   Pointer to transmit buffer (in terms of bytes).
+ * @param   rx_fr_len   Number of frames to store in recieve buffer.
+ *
+ * @return  See \ref MXC_Error_Codes for the list of error return codes.
+ */
+int MXC_SPI_TargetTransactionB(mxc_spi_regs_t *spi, uint8_t *tx_buffer, uint32_t tx_fr_len,
```
Jake-Carter, line 1056: I don't think this naming convention makes sense.

I don't see a reason to deviate from `Transaction` (blocking), `TransactionAsync` (non-blocking), and `TransactionDMA` (non-blocking).

If we want to give users a blocking/non-blocking variations of `Async` and `DMA` then that can be handled with an additional function argument.  As-is, these new functions are changing the input arguments anyways.

Libraries/PeriphDrivers/Source/SPI/spi_reva2.c
```diff
@@ -1476,23 +1526,179 @@ int MXC_SPI_RevA2_ControllerTransactionB(mxc_spi_reva_regs_t *spi, uint8_t *tx_b
                                          uint32_t tx_fr_len, uint8_t *rx_buffer, uint32_t rx_fr_len,
                                          uint8_t deassert, mxc_spi_target_t *target)
 {
```
Jake-Carter, line 1528: Maintenance overhead and code duplication for the `B` variants is a pretty major concern.  This could be handled better with an extra function argument

Jake-Carter (changes requested):

Libraries/PeriphDrivers/Include/MAX78002/spi.h
sihyung-maxim: Are you referring to SPI_SlaveTransaction instead? (Master => Controller :: Slave => Target).

sihyung-maxim: 1. Are you referring to SPI_SlaveTransaction instead? (Master => Controller :: Slave => Target). Note, the SlaveTransaction function still exists.
2. I'm doing this as a continuation from the ControllerTransaction Functions when it was decided to use this list of input parameters instead of the request struct.

Libraries/PeriphDrivers/Include/MAX78002/spi.h
sihyung-maxim: Fair. The previous names were when the Transaction functions were only interrupt-driven, and before I made the Controller/Target blocking transaction functions non-interrupt-driven. I'll change the names.

Libraries/PeriphDrivers/Include/MAX78002/spi.h
sihyung-maxim: Are we electing to use request structs for the transactions now? The ControllerTransaction functions don't use the request struct because it was previously decided that the list of parameters was preferred for the transaction functions, but not the init function.

Libraries/PeriphDrivers/Source/SPI/spi_reva2.c
sihyung-maxim: I'll add create a function similar to the previous SPI API TransSetup function to minimize duplication.

Libraries/PeriphDrivers/Include/MAX78002/spi.h
Jake-Carter: Yes, sorry.  Same point though.  `SPI_SlaveTransaction` was blocking previously

Jake-Carter: I'm getting confused about what we're trying to preserve across v1 and v2...  So the Controller/Target functions are not intended as drop-in replacements for Master/Slave?  They're just for v2?

Libraries/PeriphDrivers/Source/SPI/spi_reva2.c
sihyung-maxim: I reduced the maintenance overhead and code duplication for all the functions.

Libraries/PeriphDrivers/Include/MAX78002/spi.h
sihyung-maxim: The SPI v2 Controller/Target transaction functions weren't intended as a drop-in replacement. It was technically supposed to be a new SPI library that still supported the v1 prototypes because you preferred using a list of parameters in your fastspi library, so that's what I was referencing. However, I've updated the transaction functions to pass in request structs instead.

Libraries/PeriphDrivers/Include/MAX78002/spi.h
sihyung-maxim: I've updated the new controller/target transaction functions to follow the v1 naming convention.

Jake-Carter (changes requested):
Thanks, I think the updated transaction functions look better.

Some more generic things to address:

- [ ] : Given that the v2 changes broke some important demo projects in the last release I'd request that the old v1 API be made the default [here](https://github.com/Analog-Devices-MSDK/msdk/blob/1aaefe53063f6e009454893957b6f2677c7f9fe4/Libraries/PeriphDrivers/max78002_files.mk#L139C1-L139C18) and for all micros.  We should move towards v2 as the default in phases.

- [ ] : We should think again more carefully about whether we can achieve what we need without changing the arguments to `MXC_SPI_Init`.

  - [ ] : You've already handled the case where users _build_ with v2 but want to call `MXC_SPI_Init`.  What exactly is the difference in overhead between the `MXC_SPI_Init` and `MXC_SPI_Init_v2` in the v2 builds?  Consider universally targeting the v2 `MXC_SPI_Init` functionality for everything.  (Initialize a default struct behind the scenes?  You've added those utility functions for this...  setup over-head is almost a non-concern compared to transaction overhead)
  
  - [ ] : If the `Init_v2` implementation is _absolutely necessary_...  Given that the migration from init parameters to an init struct is a breaking change, I don't see the benefit of a new function name.  "spi.h" can be made safe against both builds by adding a compiler definition.  This is a last resort, but a clear, clean break is better than a muddy half break in my opinion.
  
    ```C
    // spi.h
    #ifdef MXC_SPI_LEGACY
    int MXC_SPI_Init(mxc_spi_regs_t *spi, int masterMode, int quadModeUsed, int numSlaves,
                     unsigned ssPolarity, unsigned int hz, mxc_spi_pins_t pins);
    #else
    int MXC_SPI_Init(mxc_spi_init_t *init);
    #endif
    ```
  
We probably have discussed the above points at some point in the past, so let's talk about it in person again since it's been some time.

USERGUIDE.md
```diff
@@ -2435,13 +2443,143 @@ For VS Code users, the `"erase flash"` [build task](#build-tasks) can be used to
...
+    :::C
+    int MXC_SPI_Init(mxc_spi_regs_t *spi, 
+                    mxc_spi_type_t controller_target, 
+                    mxc_spi_interface_t if_mode, 
+                    int numTargets, 
+                    uint8_t ts_active_pol_mask, 
+                    uint32_t freq, 
+                    mxc_spi_pins_t pins)
+
+Input Parameters:
```
Jake-Carter, line 2507: I would suggest renaming this section to "Changes:", and only highlight the things that users need to do to update their code.  This will help declutter

USERGUIDE.md
```diff
@@ -2435,13 +2443,143 @@ For VS Code users, the `"erase flash"` [build task](#build-tasks) can be used to
...
+
+- `mxc_spi_regs_t *spi` remains unchanged.
+- `int masterMode` -> `mxc_spi_type_t controller_target`. The enum `mxc_spi_type_t` was added for increased code readability.
+- `int quadModeUsed` -> `mxc_spi_interface_t if_mode`. Previously, the `MXC_SPI_Init(...)` function could only select between standard (4wire) and quad interface modes. With SPI v2, the user can select either standard (`MXC_SPI_INTERFACE_STANDARD`), quad (`MXC_SPI_INTERFACE_QUAD`), 3wire (`MXC_SPI_INTERFACE_3WIRE`), or dual (`MXC_SPI_INTERFACE_DUAL`) mode.
```
Jake-Carter, line 2511: Consider the maintenance overhead for this document when you write out the actual values of the enum here.  I think it would be better to just write "this is now selected with an enum" to keep it generic.  Intellisense and/or the Doxygen will tell users what their options are.

USERGUIDE.md
```diff
@@ -2435,13 +2443,143 @@ For VS Code users, the `"erase flash"` [build task](#build-tasks) can be used to
...
+- `int numSlaves` -> `int numTargets`. SPI v2 does not use this parameter and was kept to continue supporting SPI v1.
+- `unsigned ssPolarity` -> `uint8_t ts_active_pol_mask`. Updated to more descriptive name.
+- `unsigned int hz` -> `uint32_t freq`.
+- `mxc_spi_pins_t pins` remains unchanged.
+
+##### SPI Config Function
```
Jake-Carter, line 2517: This section describes what the SPI_Config function is, but not _how_ it's used.  Instead of listing the struct members (that's the job of the Doxygen/intellisense), a short code example of how to use the config function with relation to v1 code would be more valuable.

USERGUIDE.md
```diff
@@ -2435,13 +2443,143 @@ For VS Code users, the `"erase flash"` [build task](#build-tasks) can be used to
...
+- `mxc_spi_regs_t *spi` - Select SPI Instance to configure.
+- `mxc_spi_clkmode_t clk_mode` - Select clock mode.
+- `uint8_t frame_size` - Select single frame size (2 - 16 bits).
+- `bool use_dma_tx` - Enable SPI DMA TX (acquire and configure TX channel).
+- `bool use_dma_rx` - Enable SPI DMA RX (acquire and configure RX channel).
+- `mxc_dma_regs_t *dma` - Select DMA Instance to configure for SPI DMA (Valid only if `use_dma_tx` or `use_dma_rx` is set to true).
+
+##### SPI Transaction Functions
+
+The SPI v2 Libraries follows the terms used in the user guide: Controller and Target instead of Master and Slave, respectively.
```
Jake-Carter, line 2534: Suggest changing this to "The SPI naming convention has changed in V2 to match the terminology used by the microcontroller User Guides.  The term "Controller" is now used instead of "Master", and "Target" is used instead of "Slave".

Jake-Carter (changes requested):
@sihyung-maxim FYI I tweaked the build system to streamline the v1/v2 selection a bit.  See https://github.com/Analog-Devices-MSDK/msdk/pull/662/commits/414465974d8face20f82e041f2b44eb9b7003bae

Also made a couple minor misc. fixes.

I tested AI87 `pascalvoc-retinanetv7_3` with v1 and v2 and it's working again, which is good.  It's not a full backwards compatibility test because I mostly wrote bare-metal drivers, but it was previously broken

Just need to polish off the documentation a little more and I think this is ready

Jacob-Scheiffler (changes requested):

sihyung-maxim:
Made the changes, just retesting everything again.

Jake-Carter (approved):

Libraries/Boards/MAX78002/EvKit_V1/examples.txt
```diff
@@ -57,7 +57,8 @@ RTC_Backup
 SDHC_FAT
 SDHC_Raw
 SPI
-SPI_MasterSlave
+SPI_ControllerTarget
+SPI_MasterController
```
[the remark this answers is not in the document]
sihyung-maxim: Fixed.

Libraries/PeriphDrivers/Source/SPI/spi_ai87_v2.c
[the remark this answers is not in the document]
sihyung-maxim: Fixed.

Examples/MAX78002/SPI/README.md
```diff
@@ -5,95 +5,114 @@ MOSI (P0.21) pins.  Connect these two pins together.
...
 ### Project Usage
 
 Universal instructions on building, flashing, and debugging this project can be found in the **[MSDK User Guide](https://analog-devices-msdk.github.io/msdk/USERGUIDE/)**.
 
 ### Project-Specific Build Notes
 
-(None - this project builds as a standard example)
+Set `MXC_SPI_BUILD_V1=0` to build the SPI v2 libraries.
 
 ## Required Connections
```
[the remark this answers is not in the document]
sihyung-maxim: The pin connections are in the "Required Connections" section of the README.

Examples/MAX78002/SPI/main.c
```diff
@@ -139,79 +154,128 @@ int main(void)
...
 
-        retVal = MXC_SPI_Init_v2(&init);
+        retVal = MXC_SPI_Config(&cfg);
         if (retVal != E_NO_ERROR) {
-            printf("\nSPI INITIALIZATION ERROR\n");
+            printf("\nSPI CONFIGURATION ERROR\n");
+            return retVal;
+        }
+
+        // Custom Target Select (TS) Settings.
```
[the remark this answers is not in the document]
sihyung-maxim: Right. I removed all TS control selections in the SPI example. Moving them to SPI_ControllerTarget.

sihyung-maxim:
The SW_DRV TS control scheme made the driver too complex while still attempting to be backward compatible with SPI v1. It has lost the ease of usability at that point, so I elected to entirely remove support for that control scheme.

sihyung-maxim:
Ready for review, at the surface everything seems to be working but will do more testing in the morning just to be safe.

Jake-Carter (approved):

Jacob-Scheiffler (approved):

PR #669, opened 2023-07-13
fix(ThirdParty): Fix Wrong Output Filename for lwIP Library 

### LINKED ISSUE #665: lwIP Library: Build system generates wrong output when PROJECT is defined in a parent Makefile
I tried to build a lwIP example in VS Code and noticed the build error which states the liblwip_soft.a is not found. Dug down and found that VS Code's build task defines the `PROJECT` item while calling the makefile, lwIP's makefile tries to redefine the `PROJECT` item but it seems it can not without the `override` keyword.

Output which indicates the error:
```
  ...
  CC    C:/MaximSDK/Libraries/CMSIS/../lwIP/netif/ppp/polarssl/des.c
  CC    C:/MaximSDK/Libraries/CMSIS/../lwIP/netif/ppp/polarssl/md5.c
  CC    C:/MaximSDK/Libraries/CMSIS/../lwIP/netif/ppp/polarssl/sha1.c
  CC    C:/MaximSDK/Libraries/CMSIS/../lwIP/Maxim/mxc_eth.c
  AR    C:/MaximSDK/Libraries/lwIP/bin/softfp/lwIP_FreeRTOS.a

  arm-none-eabi-gcc.exe: error: C:/MaximSDK/Libraries/lwIP/bin/softfp/liblwip_softfp.a: No such file or directory 
```

Jake-Carter:
### Description

This PR cleans up the lwIP Makefiles to fix #665 

It also adds a missing libinfo.json for the MSDKGen project generators, which should fix missing include/browse paths for the project files.

- [x] (Optional) Link any related Github issues [using a keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)


Libraries/lwIP/lwip.mk
```diff
@@ -30,32 +30,38 @@
...
-else
-ifeq "$(MFLOAT_ABI)" "hard"
-LIBRARY_VARIANT=hardfp
+# Specify the build directory if not defined by the project
+ifeq "$(BUILD_DIR)" ""
+ifeq "$(MFLOAT_ABI)" ""
+LWIP_BUILD_DIR = ${LWIP_DIR}/bin
 else
-LIBRARY_VARIANT=softfp
+LWIP_BUILD_DIR = ${LWIP_DIR}/bin/$(MFLOAT_ABI)
```
ahmetalincak, line 39: What do you think about making default binary location to project's build directory?
```
LWIP_BUILD_DIR=$(CURDIR)/build/lwIP
```
This may also overcome some issues about library build configuration differences. You can still keep prebuild library as it is.

ahmetalincak (approved):
I've tested Makefile changes and LGTM.

Libraries/lwIP/lwip.mk
Jake-Carter: My only hesitation for doing this is that it's a somewhat larger library.  Since users will not typically modify the lwIP source code, it will affect build times.

I think there are some other solutions we can try to address the pre-built libraries in a future PR.

Jake-Carter:
> I've tested Makefile changes and LGTM.

Thanks for your testing!

ozersa:
Additionally prebuild binary shall be deleted from SDK release packet. 
During first build generating binaries for target OS will be simple and better solution instead of handling OS delta in backend.

So that these binary shall be deleted: https://github.com/Analog-Devices-MSDK/msdk/tree/main/Libraries/lwIP/bin/softfp

ozersa (approved):

Jake-Carter:
Thanks @ozersa, removed in https://github.com/Analog-Devices-MSDK/msdk/pull/669/commits/fe778bfc15ac68f157131e4872c5288c4d8e81e7

PR #675, opened 2023-07-18
fix(workflow): Fix globstar paths in labeler workflow and revise PR template

sihyung-maxim:
### Description

Fixed the globstar paths for the labeler workflow. Hopefully, it should work now.

Also, revised the PR template.


.github/labeler.yml
```diff
@@ -41,190 +41,190 @@
  ###############################################################################
 
 'BLE':
-- '*Examples/*/BLE*/*'
-- '*Libraries/BlePhy/*'
-- '*Libraries/Cordio/*'
-- '*Tools/Bluetooth/*'
+- '**/Examples/**/BLE*/**/*'
```
Jake-Carter, line 44: Doesn't make sense that we would need to _start_ this with a recursive glob.  Assuming you've already tried `Examples/**/...` and it didn't work I guess this is fine if it gets the workflow to behave.

Jake-Carter (approved):
I think the maintenance overhead of this workflow has outweighed its usefulness so far.

It will be more valuable when we start [auto-generating the release notes](https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes), but that's also something we still need to implement.

.github/labeler.yml
sihyung-maxim: Usually, there's no need to, but I noticed the location of the runner when it checks out the HEAD can sometimes vary between inside or outside the directory repo

ftariqAnalogTx (approved):

PR #679, opened 2023-07-19
feat(Examples): Add CLI Library to MSDK

Suraj-Ajjampur:
### Description

Requirements captured during functional testing of MAX78000 Examples. Ticket - [MSDK-1197](https://jira.maxim-ic.com/browse/MSDK-1197)

**Features added** 

- Asynchronous UART to read characters entered by the user
- Callback function calls a Line_Accumulator function which reads incoming characters, accumulates into a line buffer, echos characters back to the emulator, handles functions like backspace (future arrow keys, tab key etc.) and Enter key.
- Process command is called when user hits the Enter Key. This function Processes user command input into a series of tokens, all tokes are separated by whitespace characters and then it dispatches to handler functions based on the first token.
- The structure of the lookup table (command_table_t) makes it trivially easy to add a new commands to this command processor.

This implementation has been exported as a library as all clients can export this cli implementation and provide an isolation of functionality. This design is suited to be retrofitted to different CLI implementations as different arguments of a function which are separated as tokens can be passed in through the handler function.


Libraries/CLI/src/cli.c
```diff
@@ -0,0 +1,265 @@
...
+* @param void
+*
+* @return void
+*/
+void Clear_buffer(void)
+{
+    for (int i = 0; i < MAX_COMMAND_LENGTH; i++) {
+        buf[i] = '\0';
+    }
+}
```
Vignesh-VVadivel, line 239: You can use memset() function to clear the buffer instead of iterating over a for loop.

Libraries/CLI/src/cli.c
```diff
@@ -0,0 +1,265 @@
...
+
+void process_command(char *input)
+{
+    //Initialize p and end pointers
+    char *p = input;
+    char *end;
+
+    //Find end of string
+    for (end = input; *end != '\0'; end++) {}
```
Vignesh-VVadivel, line 128: One suggestion here,
Instead of using a for loop, you can use strlen(), that would give the number of characters in the input buffer. Then you can do pointer arithmetic to find the end ptr.

Libraries/CLI/src/cli.c
```diff
@@ -0,0 +1,265 @@
...
+
+/*! \name buf
+    \brief Buffer to store the characters of the command 
+*/
+char buf[256];
+int idx = 0;
+
+// Define a buffer to store commands
+#define MAX_COMMAND_LENGTH 256
+
```
Vignesh-VVadivel, line 84: You can use this MACRO to define the size of buf; like 
**char buf[MAX_COMMAND_LENGTH];**

Libraries/CLI/src/cli.c
```diff
@@ -0,0 +1,265 @@
...
+            idx--;
+            buf[idx] = '\0';
+        }
+        break;
+    }
+    case ENTER: {
+        // Handle Enter or carriage return
+        MXC_UART_WriteCharacter(MXC_UART_GET_UART(CONSOLE_UART), NEW_LINE);
+        MXC_UART_WriteCharacter(MXC_UART_GET_UART(CONSOLE_UART), ENTER);
+        buf[idx++] = '\r';
```
Vignesh-VVadivel, line 102: Do we really need a carriage return in this line?
Because I see in the process command we are looking for a NULL terminated char pointer.

Libraries/CLI/src/cli.c
```diff
@@ -0,0 +1,265 @@
...
+    bool beginning_space;
+    if (*p == SPACE || *p == TAB)
+        beginning_space = true;
+    char *argv[10];
+    int argc = 0;
+    memset(argv, 0, sizeof(argv));
+
+    //Iterate over each character in input
+    for (p = input; p < end; p++) {
+        //If in a token
```
Vignesh-VVadivel, line 141: Instead of iterating over all characters in the array, I would suggest using strchr() function, search for white_space, then create argv[][] array with all arguments given by the user. I think that could be computationally efficient.

Libraries/CLI/src/cli.c
```diff
@@ -0,0 +1,265 @@
...
+
+/* -------------------------------------------------- */
+//                      GLOBALS
+/* -------------------------------------------------- */
+
+/*! \name buf
+    \brief Buffer to store the characters of the command 
+*/
+char buf[256];
+int idx = 0;
```
Vignesh-VVadivel, line 80: One small suggestion here, we can use either uint16_t or uint32_t type for idx, since the array index cannot be negative.

Libraries/CLI/src/cli.c
```diff
@@ -0,0 +1,265 @@
...
+        buf[idx] = '\0';
+        idx = 0;
+        process_command(buf);
+        Clear_buffer();
+        break;
+    }
+
+    default: {
+        // Handle all other characters
+        if (idx < 255) {
```
Vignesh-VVadivel, line 112: We can use **if (idx < MAX_COMMAND_LENGTH-1)**

Vignesh-VVadivel (changes requested):

Examples/MAX78000/SDHC_FTHR/.vscode/settings.json
```diff
@@ -69,10 +69,20 @@
...
-        
+        "${config:board}",
+        "TARGET=${config:target}",
+        "TARGET_REV=0x4131",
+        "__GNUC__"
     ],
-    "C_Cpp.default.forcedInclude": [
-        "${workspaceFolder}/build/project_defines.h"
-    ]
+    "files.associations": {
```
Jake-Carter, line 77: Revert your changes to `.vscode/settings.json`, we don't want to track local modifications

Examples/MAX78000/SDHC_FTHR/user-cli.c
```diff
@@ -0,0 +1,141 @@
...
+void handle_unmount(int argc, char *argv[])
+{
+    umount();
+}
+/* @name handle_help
+ *
+ * @brief: Prints a help message with info about all of the supported commands.
+ *
+ */
+void handle_help(int argc, char *argv[])
```
Jake-Carter, line 136: This should be moved inside the library implementation.  The user should not have to re-implement the help function

Examples/MAX78000/SDHC_FTHR/user-cli.c
```diff
@@ -0,0 +1,141 @@
...
+    { "Unmount", handle_unmount, "Unmount card\n\r" },
+    { "Help", handle_help,
+      "Prints a help message with info about all of the supported commands.\n\r" }
+};
+
+const int num_commands = sizeof(commands) / sizeof(command_table_t);
+/*
+ * @name handle_size
+ *
+ * @brief Finds the Size of the SD Card and Free Space
```
Jake-Carter, line 56: Doxygen headers for the wrapper functions are probably unnecessary.

USERGUIDE.md
```diff
@@ -1836,3 +1836,9 @@ FreeRTOS is supported by all parts in the MSDK.  See the `FreeRTOSDemo` example
 
 - [FreeRTOS-Plus-CLI](https://www.freertos.org/FreeRTOS-Plus/index.html): **Supported**
 - [FreeRTOS-Plus-TCP](https://www.freertos.org/FreeRTOS-Plus/FreeRTOS_Plus_TCP/index.html): **Not supported** (Contributions welcome!)
+
+
+### CLI
+
+The [`Libraries/CLI`](Libraries/CLI) folder of the MSDK contains a ready to use CLI feature which uses a 2-step process of a line accumulator and a process command function to perform the CLI operationgs. Refer to [`Examples/MAX78000/SDHC_FTHR`](Examples/MAX78000/SDHC_FTHR) for implementation.
```
Jake-Carter, line 1843: This is insufficient documentation.  What we're looking for is a dedicated porting guide.  Write a new markdown file inside the `Libraries/CLI` folder with a detailed procedure on how the user is expected to add their own commands to the CLI.

Libraries/CLI/src/cli.h
```diff
@@ -0,0 +1,155 @@
...
+ */
+void process_command(char *input);
+
+/** Prints out the user command prompt sequence o the uart console 
+*
+* @param void
+*
+* @return void
+*/
+void User_Prompt_Sequence(void);
```
Jake-Carter, line 129: No need to make this a public function.  Keep it internal to the implementation file.

Libraries/CLI/src/cli.h
```diff
@@ -0,0 +1,155 @@
...
+ * @name process_command
+ *
+ * @brief Performs a Lexical analysis and tokenisis the user's commands
+ * 		  Lookup first token in a table of functions, dispatch to handler function
+ *
+ * @param   input   Character pointer containing the line accumulator input when enter key is pressed 
+ *
+ * @return  void
+ */
+void process_command(char *input);
```
Jake-Carter, line 121: I would prefer that this is also not a public function.  Instead of requiring the user to set up their UART/interrupts/callbacks to manually drive these functions, I would prefer that the user is given an `MXC_CLI_Init` function and you handle it for them behind the scenes.

Libraries/CLI/src/cli.h
```diff
@@ -0,0 +1,155 @@
...
+//                 FUNCTION PROTOTYPES
+/* -------------------------------------------------- */
+
+/**
+ * @brief Reads incoming bytes, Accumulate into line buffer, Echo's chars back to the other side and handles backspace.
+ * 		  It calls the process_command function upon pressing the ENTER key
+ *
+ * @param   user_char       User input of each character
+ */
+void line_accumlator(uint8_t user_char);
```
Jake-Carter, line 109: Typo - should be `line_accumulator`

Jake-Carter, line 109: This should also not be a public function (see `MXC_CLI_Init` comment)

Libraries/CLI/src/cli.h
```diff
@@ -0,0 +1,155 @@
...
+
+/** Prints out the user command prompt sequence o the uart console 
+*
+* @param void
+*
+* @return void
+*/
+void User_Prompt_Sequence(void);
+
+// Command table hander prototype with parameters
```
Jake-Carter, line 131: All entries in this file need [Doxygen-style](https://www.doxygen.nl/manual/docblocks.html) function headers.

Libraries/CLI/src/cli.h
```diff
@@ -0,0 +1,155 @@
...
+typedef struct
+{
+	const char *name;			/**< command string */
+	command_handler_t handler;	/**< function pointer of the handler function */
+	const char *help_string;	/**< help string of each command */
+} command_table_t;
+
+/*
+ * This table is an array of command_table_t structures that defines a set of supported commands in the program.
+ * Each command_table_t structure contains the name of the command, a function pointer to the corresponding command handler function,
```
Jake-Carter, line 144: Move this description of `command_table_t` up to the actual header for the `command_table_t` struct

Libraries/CLI/src/cli.h
```diff
@@ -0,0 +1,155 @@
...
+ * and a short description of what the command does.
+ *
+ * The structure of this lookup table makes it trivially easy to add a new command to this command processor.
```
Jake-Carter, line 147: This line is unnecessary.  Change this to inform the reader that this table is something they're are expected to define.

Libraries/CLI/src/cli.c
Suraj-Ajjampur: The line accumulator here enters a '\r' character every time the enter key is pressed by the user.
Later the process_command function looks for carriage return while iterating over each character in the input command to check the end of the string and break from the iterating for loop.

Libraries/CLI/src/cli.h
```diff
@@ -0,0 +1,155 @@
...
+ */
+extern const command_table_t commands[];
+
+//Calculates the number of commands based on commands and the command table
+extern const int num_commands;
```
Jake-Carter, line 152: Same here, inform the reader that they must provide the definition for this.  Since the `sizeof`-based definition we've implemented needs to be copy/pasted into the user's code, you can provide that to them in the comment as well.

Jake-Carter (changes requested):

Libraries/CLI/src/cli.h
Suraj-Ajjampur: Shall add this under instructions to integrate CLI section of the markdown source file.

Examples/MAX78000/SDHC_FTHR/Makefile
Suraj-Ajjampur, line 1: Default board is the FTHR_RevA for the SDHC example

sihyung-maxim (approved):

Jake-Carter (approved):
Will merge once final checks pass

PR #695, opened 2023-08-01
fix(PeriphDrivers): Increase MAX32650 I2C clock speed limitation from 400kHz to 1MHz

lorne-maxim:
The I2C driver for the MAX32650 would not allow setting the I2C clock over 400kHz even though the hardware supports up to 1MHz.  The code was modified to accept the proper range of clock settings.

### DIFF
--- a/Libraries/PeriphDrivers/Source/I2C/i2c_me10.c
+++ b/Libraries/PeriphDrivers/Source/I2C/i2c_me10.c
@@ -43,16 +43,6 @@
 #include "mxc_delay.h"
 
 /* **** Definitions **** */
-#define I2C_ERROR                                                                                \
-    (MXC_F_I2C_INT_FL0_ARB_ER | MXC_F_I2C_INT_FL0_TO_ER | MXC_F_I2C_INT_FL0_ADDR_NACK_ER |       \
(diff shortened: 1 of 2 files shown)


Libraries/PeriphDrivers/Source/I2C/i2c_me10.c
```diff
@@ -113,7 +103,7 @@ int MXC_I2C_Shutdown(mxc_i2c_regs_t *i2c)
 /* ************************************************************************** */
 int MXC_I2C_SetFrequency(mxc_i2c_regs_t *i2c, unsigned int hz)
 {
-    if (hz > MXC_I2C_FAST_SPEED) {
+    if (hz > MXC_I2C_REVB_FASTPLUS_SPEED) {
```
sihyung-maxim, line 106: This works but not sure if it's necessary to use _REVB_ definitions in the part-specific library file.

sihyung-maxim (approved):

Libraries/PeriphDrivers/Source/I2C/i2c_me10.c
Jake-Carter: Agreed, should use `MXC_I2C_FASTPLUS_SPEED` here instead

Jacob-Scheiffler:
Hey all. Since this is such a simple change and Lorne is at the career fair I went ahead and snuck this change into PR #725.

PR #715, opened 2023-08-21
fix(PeriphDrivers,Examples): Update MXC_I2C_RevX_SlaveAsyncHandler's to support I2C loopback examples at higher speeds

Jacob-Scheiffler:
One of our FAE's was inquiring about why the I2C examples did not work at higher speeds. After some digging I discovered we mishandled some of the interrupts in the MXC_I2C_RevX_SlaveAsyncHandler functions, particularly the DONE interrupt. This PR updates those handler functions to prevent them from hanging the I2C Loopback examples at higher speeds and removes the comments in the examples stating the highest speed the examples can run at is 100kHz.


Jacob-Scheiffler:
This PR's changes are captured in #719.

PR #725, opened 2023-09-06
feat(PeriphDrivers,CMSIS): Add support for multiple I2C target addresses

Jacob-Scheiffler:
This PR adds support for multiple I2C target addresses for devices which support them. For devices which support a single target address an error is returned for invalid target indices.


Jake-Carter (approved):
Looks good.

sihyung-maxim (approved):

PR #720, opened 2023-08-25
fix(SDHC)!: Fix SDHC Formats Card Size Incorrectly, Update FatFS to R0.15, Add SDXC and exFAT Support 

Jake-Carter:
#### Bug fix

The SDHC_FAT example would format cards incorrectly.  For a 32GB card, the library would format it with a card size of 1.8GB.  This was caused by an integer overflow error in the _native_ SDHC library implementation (fixed in https://github.com/Analog-Devices-MSDK/msdk/commit/9771dccdc1da6d410dd97f5bc6511fceb0d0affe)

#### FatFS Version Upgrade

As part of troubleshooting, I aded support for R0.15 of the FatFS library.  This has some potentially relevant bug fixes (see [change notes](http://elm-chan.org/fsw/ff/updates.html)).  This is marked as an API change, because FatFS R0.15 changed the input parameters for `f_mkfs` to acccept a format struct.

The default version is still R0.13, but will be updated to R0.15 next release.

Users upgrading from R0.13/R0.14 -> R0.15 can make the following change to the [f_mkfs](http://elm-chan.org/fsw/ff/00index_e.html) function.

```diff
-    if ((err = f_mkfs("", FM_ANY, 0, work, sizeof(work))) !=
+    MKFS_PARM format_options = {
+        .fmt = FM_ANY
+    };
+
+    if ((err = f_mkfs("", &format_options, work, sizeof(work))) !=
         FR_OK) {
         printf("Error formatting SD card: %s\n", FF_ERRORS[err]);
     }
```

_Alternatively_, users can set the new `FATFS_VERSION` [build configuration variable](https://analog-devices-msdk.github.io/msdk/USERGUIDE/#build-configuration-variables) to `ff13` or `ff14` to keep using the old versions.

```Makefile
# Enable SDHC library
LIB_SDHC = 1
# Use FatFS R0.13
FATFS_VERSION = ff13
```

#### SDXC and exFAT Support

We also had a feature request for SDXC card support.  This also required exFAT support to support the larger file volume sizes of SDXC.  So the ff15 configuration now enables it out of the box.  

exFAT support requires LFN (Long File-Name Support), which requires unicode support.  This increases the library size significantly.  I'm open to suggestions on how to deal with it.  Should we provide a mechanism for easily toggling SDXC support?  Or just rely on the user to edit `ffconf.h` as needed?

```shell
arm-none-eabi-size FAT32.a
   text    data     bss     dec     hex filename
[...]

PR #727, opened 2023-09-13
fix(SDHC): Fix SDHC Get Capacity and Implement Get Sectors functionality 

karaanil:
### Description

`unsigned int MXC_SDHC_Lib_GetCapacity(mxc_sdhc_csd_regs_t* csd)` function returns wrong value.
If 32 GB card is inserted to the SD card slot, `MXC_SDHC_Lib_GetCapacity` function can return at most *UINT32_MAX* as a result.
Therefore, the function prototype is changed to `unsigned long long MXC_SDHC_Lib_GetCapacity(mxc_sdhc_csd_regs_t* csd)` to get the correct result.

`unsigned int MXC_SDHC_Lib_GetSectors(mxc_sdhc_csd_regs_t* csd, int i)` function is implemented in *sdhc_lib.c* file.

### DIFF
--- a/Libraries/SDHC/Source/sdhc_lib.c
+++ b/Libraries/SDHC/Source/sdhc_lib.c
@@ -163,7 +163,7 @@ unsigned long long MXC_SDHC_Lib_GetCapacity(mxc_sdhc_csd_regs_t* csd)
 {
     unsigned int size = csd->csd.c_size;
 
-    return ((unsigned long long)size)*((unsigned long long)512*1024);
+    return ((unsigned long long)(size+1))*((unsigned long long)512*1024);
 }
 
 /* ************************************************************************** */
--- a/Libraries/SDHC/Source/sdhc_lib.c
+++ b/Libraries/SDHC/Source/sdhc_lib.c
@@ -159,11 +159,17 @@ int MXC_SDHC_Lib_GetCSD(mxc_sdhc_csd_regs_t *csd)
 }
 
 /* ************************************************************************** */
-unsigned int MXC_SDHC_Lib_GetCapacity(mxc_sdhc_csd_regs_t* csd)
+unsigned long long MXC_SDHC_Lib_GetCapacity(mxc_sdhc_csd_regs_t* csd)
 {
     unsigned int size = csd->csd.c_size;
 
-    return (size*(512*1024));
+    return ((unsigned long long)size)*((unsigned long long)512*1024);
+}
+
+/* ************************************************************************** */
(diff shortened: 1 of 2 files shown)


Libraries/SDHC/Source/sdhc_lib.c
```diff
@@ -159,11 +159,20 @@ int MXC_SDHC_Lib_GetCSD(mxc_sdhc_csd_regs_t *csd)
...
 
-    return (size*(512*1024));
+    return ((unsigned long long)size)*((unsigned long long)512*1024);
+}
+
+/* ************************************************************************** */
+unsigned int MXC_SDHC_Lib_GetSectors(mxc_sdhc_csd_regs_t* csd, int i)
+{
+    if(i == 0) {
+        return csd->csd.sector_size_0;
```
Jake-Carter, line 173: From the SD card spec, the "sector_size" field is fixed to 0x7F for SDHC and this field should not be used.

In this function we simply want to retrieve the "c_size" field to get the number of sectors on the card.  The spec indicates that it is 22 bits wide, so unsigned int should be fine as the return type.

I suggest

```C
unsigned int MXC_SDHC_Lib_GetSectors(mxc_sdhc_csd_regs_t* csd) {
    return csd->csd.c_size;
}
```

Libraries/SDHC/Source/sdhc_lib.c
```diff
@@ -159,11 +159,20 @@ int MXC_SDHC_Lib_GetCSD(mxc_sdhc_csd_regs_t *csd)
 }
 
 /* ************************************************************************** */
-unsigned int MXC_SDHC_Lib_GetCapacity(mxc_sdhc_csd_regs_t* csd)
+unsigned long long MXC_SDHC_Lib_GetCapacity(mxc_sdhc_csd_regs_t* csd)
 {
     unsigned int size = csd->csd.c_size;
 
-    return (size*(512*1024));
+    return ((unsigned long long)size)*((unsigned long long)512*1024);
```
Jake-Carter, line 166: FYI this may cause integer overflow with the FatFS library if it's casted to `DWORD`.  I fixed that overflow issue in https://github.com/Analog-Devices-MSDK/msdk/pull/720.

I think this change is ok since `MXC_SDHC_Lib_GetCapacity` is not used by FatFS directly or any existing examples.

Jake-Carter (changes requested):
I have some suggested modifications to GetSectors

Libraries/SDHC/Source/sdhc_lib.c
karaanil: changed

Jake-Carter (approved):

PR #744, opened 2023-09-25
ci(BLE, workflow): Add RSSI readings and other info in PER test

yc-adi:
Added RSSI reading in PER test and more info in the PER result plot.


EricB-ADI:
Need to add us a reviewers. Accidentally added us as assignees. Also, it says 0 files changed?

yc-adi:
Already merged before.

Issue #601, opened 2023-05-31
Verify_Register_SVD check not working correctly


jdk-maxim:
Current PR: https://github.com/Analog-Devices-MSDK/msdk/pull/600 is failing this check.

It appears to have correctly detected the affected registers and SVD files, and found one I originally missed ai85 uart_regs.h.

But the diff is failing.  Pretty sure these are correct in both the SVD and register files.  Maybe this is failing due to it being a PR from a Fork again?

Jake-Carter:
@jdk-maxim last I heard this was related to space issues on our local runners. 

@sihyung-maxim any updates on the SVD workflow lately?

Jake-Carter:
@sihyung-maxim this workflow seems to also now be broken by our move to Github Enterprise (msdk internal checkout).  See https://github.com/Analog-Devices-MSDK/msdk/actions/runs/5481437740/jobs/9985735026?pr=653

PR #751, opened 2023-10-04
feat(MiscDrivers,Examples): Add 24LC256 EEPROM driver, add I2C EEPROM example for MAX32655

kenan-balci:
Add 24LC256 EEPROM driver
Add I2C EEPROM example for MAX32655
Clang format fixes implemented.


Jake-Carter:
Thanks @kenan-balci, we will review this as soon as we can.  This week we are mostly testing our release candidate

Libraries/MiscDrivers/EEPROM/eeprom_24lc256_driver.c
```diff
@@ -0,0 +1,239 @@
...
+}
+
+/**
+ * @brief   Initializes I2C registers EEPROM
+ * @param   i2c		I2C registers
+ * @param   addr	Slave I2C address of EEPROM.
+ * @returns #E_NO_ERROR if init succeeded.
+ *
+ */
+static int eeprom_24LC256_init(mxc_i2c_regs_t *i2c, uint8_t addr)
```
Jake-Carter, line 67: Can we also move the initialization of the I2C instance inside this function?

Libraries/MiscDrivers/EEPROM/eeprom_24lc256_driver.h
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * @brief Structure with sensor function pointers
+ */
+typedef struct {
+    int (*init)(mxc_i2c_regs_t *i2c, uint8_t addr); ///< Pointer to
+    int (*read)(uint16_t addr, uint8_t *data_buffer, uint16_t length);
+    int (*write_chunk)(uint16_t addr, uint8_t *data_buffer, uint16_t length);
+    int (*write)(uint16_t addr, uint8_t *data_buffer, uint32_t length);
+} eeprom_24LC256_driver_t;
```
Jake-Carter, line 73: Doing this with a struct is interesting.  I guess it offers some more flexibility if we want to add functionality.

My only concern is that there isn't any of the usual intellisense for Doxygen headers and go-to definition lookups don't work directly.

image

I would suggest creating a typedef for the function prototypes instead of writing the prototype "inline" inside the struct.  This is similar to what we do for our callback functions (example [here](https://github.com/Analog-Devices-MSDK/msdk/blob/d6eb11cacef202f054624eab22f107264bc4bf00/Libraries/PeriphDrivers/Include/MAX78000/i2c.h#L73)).  It should let you write a Doxygen header for the function that shows up in the intellisense.

For our API functions we just want to make sure there's an explanation of what the function does, what each parameter is, and what the return value is.

Examples/MAX32655/I2C_EEPROM/.vscode/settings.json
```diff
@@ -0,0 +1,76 @@
...
+        "${config:ARM_GCC_path}/arm-none-eabi/include",
+        "${config:ARM_GCC_path}/lib/gcc/arm-none-eabi/${config:v_Arm_GCC}/include",
+        "${config:MAXIM_PATH}/Libraries/PeriphDrivers/Include/${config:target}",
+        "${config:MAXIM_PATH}/Libraries/MiscDrivers/Camera",
+        "${config:MAXIM_PATH}/Libraries/MiscDrivers/Display",
+        "${config:MAXIM_PATH}/Libraries/MiscDrivers/ExtMemory",
+        "${config:MAXIM_PATH}/Libraries/MiscDrivers/LED",
+        "${config:MAXIM_PATH}/Libraries/MiscDrivers/PMIC",
+        "${config:MAXIM_PATH}/Libraries/MiscDrivers/PushButton",
+        "${config:MAXIM_PATH}/Libraries/MiscDrivers/Touchscreen"
```
Jake-Carter, line 54: Update the MiscDrivers ["libinfo.json"](https://github.com/Analog-Devices-MSDK/msdk/blob/main/Libraries/MiscDrivers/libinfo.json) file to add the new EEPROM folder and whitelist the MAX32655.  This will get [MSDKGen](https://analog-devices-msdk.github.io/msdk/CONTRIBUTING/#contributing-examples) to automatically add the right search paths for any projects that use the drivers.

Libraries/MiscDrivers/EEPROM/eeprom_24lc256_driver.c
```diff
@@ -0,0 +1,239 @@
...
+            remaining -= I2C_MAX_READ_SIZE;
+            current_data_buffer += I2C_MAX_READ_SIZE;
+            eeprom_addr_to_read += I2C_MAX_READ_SIZE;
+        }
+    }
+    return E_NO_ERROR;
+}
+
+/**
+ * @brief   Writes a small chunk of data directly to the EEPROM. The written memory should be in the same page of EEPROM (1 page = 64 bytes)
```
Jake-Carter, line 131: Given this limitation I would suggest throwing an error code if the `length` parameter is too large.

Libraries/MiscDrivers/EEPROM/eeprom_24lc256_driver.c
```diff
@@ -0,0 +1,239 @@
...
+            send_buffer[1] = (eeprom_addr_to_read & 0x00FF);
+
+            // Read
+            err = i2c_read(&req, send_buffer, current_data_buffer, remaining);
+            if (err != E_NO_ERROR) {
+                return err;
+            } else {
+                remaining = 0;
+            }
+        } else {
```
Jake-Carter, line 113: FYI the other option we have for long transactions is interactive receive mode (IRXM).  However it's a little more CPU-intensive, since each byte needs to be processed.

Can you add a test condition to the example program for a long read with more than 256 bytes?

Jake-Carter (changes requested):
Nice work, thanks!

A few change requests below

Libraries/MiscDrivers/EEPROM/eeprom_24lc256_driver.c
kenan-balci: I moved it.

Examples/MAX32655/I2C_EEPROM/.vscode/settings.json
kenan-balci: I updated.

Libraries/MiscDrivers/EEPROM/eeprom_24lc256_driver.c
kenan-balci: I added throwing error.

Libraries/MiscDrivers/EEPROM/eeprom_24lc256_driver.c
kenan-balci: I changed the size of "Test 2" to 1024 bytes. I updated the readme file.

Libraries/MiscDrivers/EEPROM/eeprom_24lc256_driver.h
kenan-balci: I removed the structure. I wrote the function prototypes with Doxygen headers. I added a "_eeprom_24lc256_req_t" struct to the header that keeps EEPROM's I2C information similar to "_i2c_req_t". [here](https://github.com/Analog-Devices-MSDK/msdk/blob/d6eb11cacef202f054624eab22f107264bc4bf00/Libraries/PeriphDrivers/Include/MAX78000/i2c.h#L103C15-L103C15)
I didn't add the callback functions. Is this required?

Examples/MAX32655/I2C_EEPROM/main.c
```diff
@@ -73,28 +76,23 @@ int main(void)
     printf("Press ENTER key to Continue\n\n");
     getchar();
 
-    err = MXC_I2C_Init(I2C_MASTER, 1, 0);
+    err = Eeprom_24LC256_Init(&eeprom1_req, I2C_MASTER, EEPROM_24LC256_I2C_SLAVE_ADDR0,
```
Jake-Carter, line 79: Thanks @kenan-balci, this looks good.  I have one more design suggestion:

- Remove the `&eeprom1_req` as a required parameter for the functions.

You can move the `eeprom_24lc256_req_t` struct to be _internal_ to the drivers so that the user doesn't need to pass the pointer in every time.

Since the user will not be interacting with this struct directly there's no need to expose it _unless_ you want to preserve the ability to interface with multiple EEPROMs at the same time.  If that's the case then you can leave it.

Examples/MAX32655/I2C_EEPROM/main.c
```diff
@@ -73,28 +76,23 @@ int main(void)
...
-        err = eeprom_24LC256.write(eeprom_memory_addr, &written_val, 1);
+        err = Eeprom_24LC256_Write(&eeprom1_req, eeprom_memory_addr, &written_val, 1);
         if (err != E_NO_ERROR) {
             printf("EEPROM write error, error code = %d\n", err);
         } else {
             printf("The value: 0x%02X is written to the address: 0x%04X\n", written_val,
                    eeprom_memory_addr);
         }
-        err = eeprom_24LC256.read(eeprom_memory_addr, &readed_val, 1);
+        err = Eeprom_24LC256_Read(&eeprom1_req, eeprom_memory_addr, &readed_val, 1);
```
Jake-Carter, line 95: Minor typo - instead of `readed_val` use `read_val`

Jake-Carter (changes requested):

Libraries/MiscDrivers/EEPROM/eeprom_24lc256_driver.h
Jake-Carter: Thanks.  No, the callback isn't required since you're using the synchronous API functions.

One more suggestion - see https://github.com/Analog-Devices-MSDK/msdk/pull/751#discussion_r1372161562

Jake-Carter:
Seems to be suffering from https://github.com/Analog-Devices-MSDK/msdk/issues/601

Verify Register workflow can be ignored

Examples/MAX32655/I2C_EEPROM/main.c
kenan-balci: I changed.

Examples/MAX32655/I2C_EEPROM/main.c
kenan-balci: My purpose is to give the driver the ability to communicate with more than one EEPROM, as you said. Up to 8 EEPROMs with different slave addresses can be connected to the same I2C bus. I tested communication with two EEPROMs, and the driver worked properly.

Jake-Carter (approved):

Jake-Carter:
@sihyung-maxim / @Jacob-Scheiffler please take a look at this when you have the chance

lorne-maxim (approved):
@sihyung-maxim, the new files in this PR will need to be updated with our latest copyrights.

Jacob-Scheiffler (approved):

PR #754, opened 2023-10-05
fix(ThirdParty,Examples): Add missing dependencies to LVGL makefiles

lorne-maxim:
Add the missing statements needed for all files in the Libraries/LVGL/lvgl/src/extra directories.
Disable Flash ECC in the MAX32672 Flash example.  It uses 32-bit writes which will not work when ECC is enabled.


Jake-Carter (changes requested):
I'd suggest using `IPATH` instead of `CFLAGS += -I...`

lorne-maxim:
I thought it was strange that the other makefiles weren't using IPATH.  I just stuck with the style used by the others, though.  What do you think?  Should we rework all the makefiles in LVGL?

Jake-Carter:
Eh...  at this point that's probably a better cleanup task for a later date.  If they all use this convention this'll work

Jake-Carter (approved):

lorne-maxim:
That's what I was hoping you'd say!

PR #779, opened 2023-10-24
fix(PeriphDrivers): Added update to force scale select sample to occur with ADC and ADC RefSelect for REVA API

EricB-ADI:
### Description
For chips with an ADC which allows for external voltage division, the scale select is only sampled after the channel select is switched to a new channel. So when using channel one, the external scale will not take effect since 0 is the default channel and is not switched even if configured. This fix forcefully toggles the channel selected anytime the ExtScale is set.

Also added is the API for ADC RefSelect for REVA

### DIFF
--- a/Libraries/CMSIS/Device/Maxim/MAX32665/Include/adc_regs.h
+++ b/Libraries/CMSIS/Device/Maxim/MAX32665/Include/adc_regs.h
@@ -6,6 +6,23 @@
  */
 
 /******************************************************************************
+ *
+ * Copyright 2023 Analog Devices, Inc.
+ *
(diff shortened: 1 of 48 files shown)


Jake-Carter (approved):
Looks good, just need to add `mxc_adc_ref_t` to MAX32650 and MAX32665 top-level `adc.h` headers to fix the build errors

Jake-Carter:
Bot updated headers :/

https://github.com/Analog-Devices-MSDK/msdk/pull/779/commits/c8b648e2bd0e84385dd13bf8d0c2ebbc3feb3507

EdwinFairchild (approved):

PR #805, opened 2023-11-21
fix(BLE): Connection-Parameter-Update

EricB-ADI:
### Description
Cordio now will only cancel connection parameter update requests upon max attempts or connection interval agreement.

At the app layer, when requesting a connection update, a request goes through periodically until the update is successful or the max attempts have been reached. However, the stack does not differentiate between Master and Slave requests. So if the Master requests a connection update, the slaves request will be cancelled. Therefore, the parameters the application wants are never obtained. This change will stop trying for updates once the connection interval is accepted based on the range given by the user, or max attempts have been reached.


EdwinFairchild (approved):

AbbyWolf-ADI (approved):

EdwinFairchild:
Let the record show Eric "verified" this, I was paid to approve.

PR #806, opened 2023-11-22
fix(Examples): Fixed bug in FaceId RISC-V example

vicloginov:
Fixed bug of FaceId RISC-V example when it's executed on MAX78000 Feather board.

### DIFF
--- a/Libraries/Cordio/ble-profiles/sources/af/app_slave.c
+++ b/Libraries/Cordio/ble-profiles/sources/af/app_slave.c
@@ -523,9 +523,13 @@ static void appSlaveConnUpdate(dmEvt_t *pMsg, appConnCb_t *pCb)
 {
   if (pAppUpdateCfg->idlePeriod != 0)
   {
-    /* if successful */
-    if (pMsg->hdr.status == HCI_SUCCESS)
+    const bool_t intervalInBetween =  pAppUpdateCfg->connIntervalMin <= pMsg->connUpdate.connInterval && 
+                                      pMsg->connUpdate.connInterval <= pAppUpdateCfg->connIntervalMax;
+    
+    /* if successful and we got an update using one of our requested values*/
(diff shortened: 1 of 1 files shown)


Examples/MAX78000/CNN/faceid_evkit-riscv/project.mk
```diff
@@ -23,7 +23,7 @@ endif
 # Place build files specific to FTHR_RevA here.
 ifeq "$(BOARD)" "FTHR_RevA"
 # Only Enable if 2.4" TFT is connected to Feather
-#PROJ_CFLAGS+=-DENABLE_TFT
+#PROJ_CFLAGS+=-DTFT_ENABLE
```
Jake-Carter, line 26: Mismatch between `TFT_ENABLE` here

Examples/MAX78000/CNN/kws20_demo-riscv/project.mk
```diff
@@ -19,7 +20,7 @@ endif
 # Place build files specific to FTHR_RevA here.
 ifeq "$(BOARD)" "FTHR_RevA"
 # Only Enable if 2.4" TFT is connected to Feather
-# PROJ_CFLAGS+=-DENABLE_TFT
+#PROJ_CFLAGS+=-DENABLE_TFT
```
Jake-Carter, line 23: and `ENABLE_TFT` here

Jake-Carter (changes requested):
Thanks @vicloginov

One minor comment below - can we use `TFT_ENABLE` or `ENABLE_TFT` across both examples?  I think it will help if we make this definition consistent

Examples/MAX78000/CNN/kws20_demo-riscv/project.mk
vicloginov: Agree. I will rename it.

PR #807, opened 2023-11-22
fix(Examples): Fixed bug in FaceId RISC-V example

vicloginov:
Fixed bug of FaceID RISC-V example when it's executed on MAX78000 Feather board.
Use TFT_ENABLE setting for all CNN examples.


Jake-Carter (approved):
Thanks Victor!

Jacob-Scheiffler (approved):

PR #808, opened 2023-11-24
feat(PeriphDrivers): Add clock selection feature to WDT

hfakkiz:
### Description

WDT peripheral of MAX32655, MAX32662, MAX32670, MAX32672, MAX32675, MAX32680, MAX32690, MAX78000 and MAX78002 supports selecting different clock sources which are given in user guide. 
Added a new function to set clksel register on revB WDT driver.
Added a new function to select supported clock sources for these part numbers.

### DIFF
--- a/Libraries/PeriphDrivers/Include/MAX32655/wdt.h
+++ b/Libraries/PeriphDrivers/Include/MAX32655/wdt.h
@@ -121,6 +121,16 @@ typedef enum {
 } mxc_wdt_mode_t;
 
 /**
+ * @brief Peripheral Clock settings 
+ */
+typedef enum {
+    MXC_WDT_PCLK = 0,
+    MXC_WDT_IBRO_CLK,
+    MXC_WDT_INRO_CLK,
+    MXC_WDT_ERTCO_CLK
+} mxc_wdt_clock_t;
+
+/**
  * @brief Timer Configuration
  */
 typedef struct {
@@ -233,6 +243,13 @@ int MXC_WDT_GetIntFlag(mxc_wdt_regs_t *wdt);
  * @param       wdt     Pointer to watchdog registers.
  */
 void MXC_WDT_ClearIntFlag(mxc_wdt_regs_t *wdt);
+
+/**
+ * @brief       Sets clock source.
+ * @param       wdt             Pointer to watchdog registers.
+ * @param       clock_source    Clock source.
+ */
+int MXC_WDT_SetClockSource(mxc_wdt_regs_t *wdt, mxc_wdt_clock_t clock_source);
 
 /**@} end of group wdt */
 
--- a/Libraries/PeriphDrivers/Source/WDT/wdt_ai87.c
+++ b/Libraries/PeriphDrivers/Source/WDT/wdt_ai87.c
@@ -157,3 +157,38 @@ void MXC_WDT_ClearIntFlag(mxc_wdt_regs_t *wdt)
 {
     MXC_WDT_RevB_ClearIntFlag((mxc_wdt_revb_regs_t *)wdt);
 }
+
+int MXC_WDT_SetClockSource(mxc_wdt_regs_t *wdt, mxc_wdt_clock_t clock_source)
+{
+    const uint8_t clock_source_num = 8;
+    uint8_t idx = 0;
+    uint8_t instance = 0;
+    mxc_wdt_clock_t clock_sources[2][8] = {
+        { MXC_WDT_PCLK, MXC_WDT_IPO_CLK, MXC_WDT_IBRO_CLK, MXC_WDT_INRO_CLK, MXC_WDT_ERTCO_CLK,
+          MXC_WDT_EXT_CLK, MXC_WDT_ERFO_CLK, 0xFF },
+        { MXC_WDT_PCLK, MXC_WDT_IPO_CLK, MXC_WDT_IBRO_CLK, MXC_WDT_INRO_CLK, MXC_WDT_ERTCO_CLK,
+          MXC_WDT_EXT_CLK, MXC_WDT_ERFO_CLK, 0xFF }
+    };
+
+    if (wdt == MXC_WDT0) {
+        instance = 0;
+    } else if (wdt == MXC_WDT1) {
+        instance = 1;
+    } else {
+        return E_BAD_PARAM;
+    }
+
+    for (idx = 0; idx < clock_source_num; idx++) {
+        if (clock_sources[instance][idx] == clock_source) {
+            break;
+        }
+    }
+
+    if (idx == clock_source_num) {
+        return E_BAD_PARAM;
+    }
+
+    MXC_WDT_RevB_SetClockSource((mxc_wdt_revb_regs_t *)wdt, idx);
+
+    return E_NO_ERROR;
+}
--- a/Libraries/PeriphDrivers/Source/WDT/wdt_ai87.c
+++ b/Libraries/PeriphDrivers/Source/WDT/wdt_ai87.c
@@ -160,12 +160,6 @@ void MXC_WDT_ClearIntFlag(mxc_wdt_regs_t *wdt)
 
 int MXC_WDT_SetClockSource(mxc_wdt_regs_t *wdt, mxc_wdt_clock_t clock_source)
 {
-    const uint8_t last_clock_source = MXC_WDT_ERFO_CLK;
-
-    if ((clock_source < 0) || (clock_source > last_clock_source)) {
-        return E_BAD_PARAM;
-    }
-
     MXC_WDT_RevB_SetClockSource((mxc_wdt_revb_regs_t *)wdt, (int)clock_source);
 
     return E_NO_ERROR;
--- a/Libraries/PeriphDrivers/Source/WDT/wdt_ai87.c
+++ b/Libraries/PeriphDrivers/Source/WDT/wdt_ai87.c
@@ -160,35 +160,13 @@ void MXC_WDT_ClearIntFlag(mxc_wdt_regs_t *wdt)
 
 int MXC_WDT_SetClockSource(mxc_wdt_regs_t *wdt, mxc_wdt_clock_t clock_source)
 {
-    const uint8_t clock_source_num = 8;
-    uint8_t idx = 0;
-    uint8_t instance = 0;
-    mxc_wdt_clock_t clock_sources[2][8] = {
-        { MXC_WDT_PCLK, MXC_WDT_IPO_CLK, MXC_WDT_IBRO_CLK, MXC_WDT_INRO_CLK, MXC_WDT_ERTCO_CLK,
-          MXC_WDT_EXT_CLK, MXC_WDT_ERFO_CLK, 0xFF },
-        { MXC_WDT_PCLK, MXC_WDT_IPO_CLK, MXC_WDT_IBRO_CLK, MXC_WDT_INRO_CLK, MXC_WDT_ERTCO_CLK,
-          MXC_WDT_EXT_CLK, MXC_WDT_ERFO_CLK, 0xFF }
-    };
+    const uint8_t last_clock_source = MXC_WDT_ERFO_CLK;
 
-    if (wdt == MXC_WDT0) {
-        instance = 0;
-    } else if (wdt == MXC_WDT1) {
-        instance = 1;
-    } else {
+    if ((clock_source < 0) || (clock_source > last_clock_source)) {
         return E_BAD_PARAM;
     }
 
-    for (idx = 0; idx < clock_source_num; idx++) {
-        if (clock_sources[instance][idx] == clock_source) {
-            break;
-        }
-    }
-
-    if (idx == clock_source_num) {
-        return E_BAD_PARAM;
-    }
-
-    MXC_WDT_RevB_SetClockSource((mxc_wdt_revb_regs_t *)wdt, idx);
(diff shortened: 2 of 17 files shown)


Libraries/PeriphDrivers/Source/WDT/wdt_ai87.c
```diff
@@ -157,3 +157,38 @@ void MXC_WDT_ClearIntFlag(mxc_wdt_regs_t *wdt)
 {
     MXC_WDT_RevB_ClearIntFlag((mxc_wdt_revb_regs_t *)wdt);
 }
+
+int MXC_WDT_SetClockSource(mxc_wdt_regs_t *wdt, mxc_wdt_clock_t clock_source)
+{
+    const uint8_t clock_source_num = 8;
+    uint8_t idx = 0;
+    uint8_t instance = 0;
+    mxc_wdt_clock_t clock_sources[2][8] = {
```
Jake-Carter, line 166: For micros where all WDT instances have the same available clock sources you can reduce this

```C
int MXC_WDT_SetClockSource(mxc_wdt_regs_t *wdt, mxc_wdt_clock_t clock_source)
{
    const uint8_t clock_source_num = 8;
    uint8_t idx = 0;
    uint8_t instance = 0;
    mxc_wdt_clock_t clock_sources[2][8] = {
        { MXC_WDT_PCLK, MXC_WDT_IPO_CLK, MXC_WDT_IBRO_CLK, MXC_WDT_INRO_CLK, MXC_WDT_ERTCO_CLK,
          MXC_WDT_EXT_CLK, MXC_WDT_ERFO_CLK, 0xFF },
        { MXC_WDT_PCLK, MXC_WDT_IPO_CLK, MXC_WDT_IBRO_CLK, MXC_WDT_INRO_CLK, MXC_WDT_ERTCO_CLK,
          MXC_WDT_EXT_CLK, MXC_WDT_ERFO_CLK, 0xFF }
    };

    if (wdt == MXC_WDT0) {
        instance = 0;
    } else if (wdt == MXC_WDT1) {
        instance = 1;
    } else {
        return E_BAD_PARAM;
    }

    for (idx = 0; idx < clock_source_num; idx++) {
        if (clock_sources[instance][idx] == clock_source) {
            break;
        }
    }

    if (idx == clock_source_num) {
        return E_BAD_PARAM;
    }

    MXC_WDT_RevB_SetClockSource((mxc_wdt_revb_regs_t *)wdt, idx);

    return E_NO_ERROR;
}
```

to this.  It will help reduce unnecessary code size.  

I'd also suggest adding a note to the `mxc_wdt_clock_t` definition that the ordering of the enum should match the register field exactly.

```C
int MXC_WDT_SetClockSource(mxc_wdt_regs_t *wdt, mxc_wdt_clock_t clock_source)
{
    MXC_WDT_RevB_SetClockSource((mxc_wdt_revb_regs_t *)wdt, (int)clock_source);

    return E_NO_ERROR;
}
```

Libraries/PeriphDrivers/Source/WDT/wdt_me12.c
```diff
@@ -163,3 +163,30 @@ void MXC_WDT_ClearIntFlag(mxc_wdt_regs_t *wdt)
 {
     MXC_WDT_RevB_ClearIntFlag((mxc_wdt_revb_regs_t *)wdt);
 }
+
+int MXC_WDT_SetClockSource(mxc_wdt_regs_t *wdt, mxc_wdt_clock_t clock_source)
+{
+    const uint8_t clock_source_num = 8;
+    uint8_t idx = 0;
+    mxc_wdt_clock_t clock_sources[1][8] = { { MXC_WDT_PCLK, MXC_WDT_IPO_CLK, MXC_WDT_IBRO_CLK,
```
Jake-Carter, line 171: You could also apply the same reduction here, since there is only a single WDT instance

Libraries/PeriphDrivers/Source/WDT/wdt_me17.c
```diff
@@ -153,3 +153,47 @@ void MXC_WDT_ClearIntFlag(mxc_wdt_regs_t *wdt)
...
+        { MXC_WDT_PCLK, MXC_WDT_IBRO_CLK, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF },
+        { MXC_WDT_IBRO_CLK, MXC_WDT_ERTCO_CLK, MXC_WDT_INRO_CLK, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF }
+    };
+#elif TARGET_NUM == 32680
+    mxc_wdt_clock_t clock_sources[2][8] = {
+        { MXC_WDT_PCLK, MXC_WDT_IBRO_CLK, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF },
+        { MXC_WDT_IBRO_CLK, 0xFF, MXC_WDT_INRO_CLK, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF }
+    };
+#else
+    mxc_wdt_clock_t clock_sources[2][8] = { { 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF },
```
Jake-Carter, line 174: Throw a build error here instead with `#error ...`.  This will help us catch cases where this file is reused in the future.

Libraries/PeriphDrivers/Source/WDT/wdt_me18.c
```diff
@@ -153,3 +153,36 @@ void MXC_WDT_ClearIntFlag(mxc_wdt_regs_t *wdt)
 {
     MXC_WDT_RevB_ClearIntFlag((mxc_wdt_revb_regs_t *)wdt);
 }
+
+int MXC_WDT_SetClockSource(mxc_wdt_regs_t *wdt, mxc_wdt_clock_t clock_source)
+{
+    const uint8_t clock_source_num = 8;
```
Jake-Carter, line 159: Why's `clock_source_num` always 8?

Libraries/PeriphDrivers/Include/MAX32655/wdt.h
```diff
@@ -120,6 +120,16 @@ typedef enum {
...
 
+/**
+ * @brief Peripheral Clock settings 
+ */
+typedef enum {
+    MXC_WDT_PCLK = 0,
+    MXC_WDT_IBRO_CLK,
+    MXC_WDT_INRO_CLK,
+    MXC_WDT_ERTCO_CLK
+} mxc_wdt_clock_t;
```
Jake-Carter, line 131: It'd be ideal if we didn't need to create this, and could just use `mxc_sys_system_clock_t` as the second parameter to the function.

```C
/**
 * @brief       Sets clock source.
 * @param       wdt             Pointer to watchdog registers.
 * @param       clock_source    Clock source.
 */
int MXC_WDT_SetClockSource(mxc_wdt_regs_t *wdt, mxc_system_clock_t clock_source);
```

The default PCLK option is problematic though...  no action needed on this unless you have any ideas.  The less enums we have to maintain that are specific to each micro the better

Jake-Carter (changes requested):
Thanks @hfakkiz, looks good.  A couple minor comments below

Libraries/PeriphDrivers/Source/WDT/wdt_me18.c
hfakkiz: WDT supports 8 different clock sources but mostly not using all of 8 options. I set `clock_source_num` parameter and `clock_sources` array size to 8 because upcoming revisions can use other options which are not used now. To reduce code size, I can change them to number of supported clock source.

Libraries/PeriphDrivers/Source/WDT/wdt_me17.c
hfakkiz: Updated code for ME15 and ME17.

Libraries/PeriphDrivers/Source/WDT/wdt_ai87.c
hfakkiz: Updated code for ME12, ME15, ME21 and AI87.

Libraries/PeriphDrivers/Include/MAX32655/wdt.h
hfakkiz: As you said PCLK option is a problem for this solution. Also, using `mxc_wdt_clock_t` enum helps to use only supported clock sources of WDT. If it's not a problem, I prefer to use `mxc_wdt_clock_t` enum solution.

Libraries/PeriphDrivers/Source/WDT/wdt_me18.c
```diff
@@ -153,3 +153,36 @@ void MXC_WDT_ClearIntFlag(mxc_wdt_regs_t *wdt)
...
+    uint8_t idx = 0;
+    uint8_t instance = 0;
+    mxc_wdt_clock_t clock_sources[2][8] = {
```
Jake-Carter, line 162: Initialize with `clock_source_num` instead of `8`

```C
mxc_wdt_clock_t clock_sources[2][clock_source_num] = ...
```

Jake-Carter (changes requested):
Thank you @hfakkiz, one last change request based on your latest comments

Libraries/PeriphDrivers/Source/WDT/wdt_revb.c
```diff
@@ -154,3 +154,9 @@ void MXC_WDT_RevB_ClearIntFlag(mxc_wdt_revb_regs_t *wdt)
 {
     wdt->ctrl &= ~(MXC_F_WDT_REVB_CTRL_INT_LATE | MXC_F_WDT_REVB_CTRL_INT_EARLY);
 }
+
+void MXC_WDT_RevB_SetClockSource(mxc_wdt_revb_regs_t *wdt, int clock_source)
+{
```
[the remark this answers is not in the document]
hfakkiz: Done

Libraries/PeriphDrivers/Source/WDT/wdt_me18.c
Jake-Carter: Hmm strange...  should work as long as it's `const`.  What's the error?

hfakkiz: It is "error: variable-sized object may not be initialized".

image

Jake-Carter:
Thanks @hfakkiz, looks good to go.  We'll get a second review then merge.

Libraries/PeriphDrivers/Source/WDT/wdt_me18.c
```diff
@@ -153,3 +153,32 @@ void MXC_WDT_ClearIntFlag(mxc_wdt_regs_t *wdt)
...
+        return E_BAD_PARAM;
+    }
+
+    for (idx = 0; idx < clock_source_num; idx++) {
+        if (clock_sources[instance][idx] == clock_source) {
+            break;
+        }
+    }
+
+    MXC_WDT_RevB_SetClockSource((mxc_wdt_revb_regs_t *)wdt, idx);
```
ozersa, line 181: It seems it will be nice if we have some utilities functions place, for this case, if we have a function like below in some place it will simplify me17/me18... source codes.
`
int mxc_util_get_index(int item, int *arr, int num_of_item)
{
    int i;

    for (i = (num_of_item-1); i >= 0; i--) {
        if (arr[i] == item) {
            break;
        }
    }
    
    return i; // on success return index, on fail return -1;
}
`

Libraries/PeriphDrivers/Source/WDT/wdt_revb.c
```diff
@@ -154,3 +154,12 @@ void MXC_WDT_RevB_ClearIntFlag(mxc_wdt_revb_regs_t *wdt)
...
+    const uint8_t clock_source_num = 8; // Max number of clock sources for Rev B WDT
+
+    MXC_ASSERT(clock_source < clock_source_num && clock_source >= 0);
```
ozersa, line 162: Please use parenthesis

Libraries/PeriphDrivers/Source/WDT/wdt_me18.c
```diff
@@ -153,3 +153,32 @@ void MXC_WDT_ClearIntFlag(mxc_wdt_regs_t *wdt)
...
+        instance = 1;
+    } else {
+        return E_BAD_PARAM;
+    }
+
+    for (idx = 0; idx < clock_source_num; idx++) {
+        if (clock_sources[instance][idx] == clock_source) {
+            break;
+        }
+    }
```
ozersa, line 179: It seems it will be nice if we have some utilities functions place, for this case, if we have a function like below in some place it will simplify me17/me18... source codes.
`
int mxc_util_get_index(int item, int *arr, int num_of_item)
{
    int i;

    for (i = (num_of_item-1); i >= 0; i--) {
        if (arr[i] == item) {
            break;
        }
    }
    
    return i; // on success return index, on fail return -1;
}
`

hfakkiz: For ME17 and ME18, `clock_sources` arrays have 2 or 3 elements from the beginning of arrays. For loop solution would find index in maximum 3 iterations. Using a search function like `lfind(...)` instead of this for loop will make it much more complicated.

Libraries/PeriphDrivers/Source/WDT/wdt_revb.c
hfakkiz: Done.

Jacob-Scheiffler (approved):

ozersa (approved):

PR #818, opened 2023-12-06
fix(PeriphDrivers): Fix Stuck Timer When PWM is set to 0

ozersa:
### Description

For PWM mode when timer period end the tmr->count is reset to 0x00000001 and the timer resumes counting. 
On edge case (pwm==0) flow stuck in while loop, tmr->cnt newer become bigger than pwm so that +8 added.

- [X] PR Title follows correct guidelines.
- [X] Description of changes and all other relevant information.
- [X] (Optional) Provide info on any relevant functional testing/validation.  For API changes or significant features, this is not optional.

Test detail

If tolerance value be +8
image

If tolerance value be +2
image

As can be seen below due to if checks it is hard to catch exact 0x000001 value, there should be a tolerance value, so that i set it as 8 (TMR clock is PCLK,   prescaler=1)
image

### DIFF
--- a/Libraries/PeriphDrivers/Source/TMR/tmr_revb.c
+++ b/Libraries/PeriphDrivers/Source/TMR/tmr_revb.c
@@ -269,10 +269,28 @@ int MXC_TMR_RevB_SetPWM(mxc_tmr_revb_regs_t *tmr, uint32_t pwm)
         return E_BAD_PARAM;
     }
 
-    while (tmr->cnt >= pwm) {}
+    bool timera_is_running = tmr->ctrl0 & MXC_F_TMR_CTRL0_EN_A;
+    bool timerb_is_running = tmr->ctrl0 & MXC_F_TMR_CTRL0_EN_B;
+
+    if (timera_is_running || timerb_is_running) {
+        MXC_TMR_RevB_ClearFlags(tmr); // Clear flags so we can catch the next one
+        while (!MXC_TMR_RevB_GetFlags(tmr)) {} // Wait for next PWM transition
+        MXC_TMR_RevB_Stop(tmr); // Pause timer
+        MXC_TMR_RevB_SetCount(tmr, 0); // Reset the count
+        MXC_TMR_RevB_ClearFlags(
+            tmr); // Clear flags since app code wants the new PWM transitions set by this function
+    }
 
     tmr->pwm = pwm;
     while (!(tmr->intfl & MXC_F_TMR_REVB_INTFL_WRDONE_A)) {}
+
+    if (timera_is_running) {
+        tmr->ctrl0 |= MXC_F_TMR_REVB_CTRL0_EN_A; // Unpause A
+    }
+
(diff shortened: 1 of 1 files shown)


Libraries/PeriphDrivers/Source/TMR/tmr_revb.c
```diff
@@ -269,7 +269,16 @@ int MXC_TMR_RevB_SetPWM(mxc_tmr_revb_regs_t *tmr, uint32_t pwm)
...
+    /*
+     * when timer period end the tmr->count is reset to 0x00000001 
+     * and the timer resumes counting. 
+     * On edge case when pwm==0 flow stuck in below code due to
+     * tmr->cnt newer become bigger than pwm.
+     * So that at least +2 shall be added.
+     * But there should be a tolerance too, to it be catched
+     * So that +8 added as toleransa value.
+     */
+    while (tmr->cnt >= (pwm + 8)) {}
```
Jake-Carter, line 281: @ozersa  What about monitoring the interrupt flag set by the hardware?

image

We only need to do this check if the timer is currently running.  Something like:

```C
int MXC_TMR_RevB_SetPWM(mxc_tmr_revb_regs_t *tmr, uint32_t pwm)
{
    int tmr_id = MXC_TMR_GET_IDX((mxc_tmr_regs_t *)tmr);
    (void)tmr_id;
    MXC_ASSERT(tmr_id >= 0);

    if (pwm > (tmr->cmp)) {
        return E_BAD_PARAM;
    }

    bool timera_is_running = tmr->ctrl0 & MXC_F_TMR_CTRL0_EN_A;
    bool timerb_is_running = tmr->ctrl0 & MXC_F_TMR_CTRL0_EN_B;

    if (timera_is_running || timerb_is_running) {
        MXC_TMR_RevB_ClearFlags(tmr); // Clear flags so we can catch the next one
        while (!MXC_TMR_RevB_GetFlags(tmr)) {} // Wait for next PWM transition
        MXC_TMR_RevB_Stop(tmr); // Pause timer
        MXC_TMR_SetCount(tmr, 0); // Reset the count
        MXC_TMR_ClearFlags(tmr); // Clear flags since app code wants the new PWM transitions set by this function
    }

    tmr->pwm = pwm;
    while (!(tmr->intfl & MXC_F_TMR_REVB_INTFL_WRDONE_A)) {}

    if (timera_is_running) {
        tmr->ctrl0 |= MXC_F_TMR_REVB_CTRL0_EN_A; // Unpause A
    }

    if (timerb_is_running) {
        tmr->ctrl0 |= MXC_F_TMR_REVB_CTRL0_EN_A; // Unpause B
    }

    return E_NO_ERROR;
}
```

ozersa: The solution works. 
Commit updated.

Jake-Carter (approved):

sihyung-maxim (approved):

PR #819, opened 2023-12-06
fix(CMSIS): Vector table address now properly set

EricB-ADI:
### Description
Inside the startup script for the RISCV MAX32655, the vector table was being set to 0, instead of the the ISR vector table so interrupts were not working correctly.

### DIFF
--- a/Libraries/CMSIS/5.9.0/Core/Include/core_rv32.h
+++ b/Libraries/CMSIS/5.9.0/Core/Include/core_rv32.h
@@ -484,7 +484,13 @@ __STATIC_INLINE uint32_t NVIC_GetPendingIRQ(IRQn_Type IRQn)
  */
 __STATIC_INLINE void NVIC_SetPendingIRQ(IRQn_Type IRQn)
 {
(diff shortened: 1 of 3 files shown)


Jake-Carter:
@EricB-ADI I see on the AI87 we're doing

```C
Reset_Handler:

  /* set start of flash in mvtec, with vectored mode */
  la    t0, (__FlashStart + 1)
  csrrw x0, mtvec, t0
```

I think we should also update this to use `__isr_vector` instead?

EricB-ADI:
> @EricB-ADI I see on the AI87 we're doing
> 
> ```c
> Reset_Handler:
> 
>   /* set start of flash in mvtec, with vectored mode */
>   la    t0, (__FlashStart + 1)
>   csrrw x0, mtvec, t0
> ```
> 
> I think we should also update this to use `__isr_vector` instead?

I dont think it makes a difference because the ISR vector table is the first thing placed into the linker script of the text section. So they are the same. However, I think it is more explicit what is happening if you spell out isr vector/

PR #834, opened 2023-12-14
fix(PeriphDrivers): Fix TMR Shutdown Bug, Stop Timer Before Clearing CTRL0

hfakkiz:
### Description

- In shutdown function, ctrl0 register set as 0. This causes problem because it disables timer clock before the timer. Thats why cnt register starts counting in next timer initialization when clk_en_x field is enabled even en_x field read as 0. To fix this problem, called stop function before setting ctrl0 register to 0.

image

For these 2 programs, timer shutdown and reinitialize for each configuration. Output 1 is current revB driver's output and output 2 is updated version's output. 

### DIFF
--- a/Libraries/PeriphDrivers/Source/TMR/tmr_revb.c
+++ b/Libraries/PeriphDrivers/Source/TMR/tmr_revb.c
@@ -234,6 +234,8 @@ void MXC_TMR_RevB_Shutdown(mxc_tmr_revb_regs_t *tmr)
     (void)tmr_id;
     MXC_ASSERT(tmr_id >= 0);
 
+    // Stop timer before disable it.
+    MXC_TMR_RevB_Stop(tmr);
     // Disable timer and clear settings
     tmr->ctrl0 = 0;
     while (tmr->ctrl1 & MXC_F_TMR_REVB_CTRL1_CLKRDY_A) {}


ozersa (approved):

Libraries/PeriphDrivers/Source/TMR/tmr_revb.c
```diff
@@ -276,12 +278,12 @@ int MXC_TMR_RevB_SetPWM(mxc_tmr_revb_regs_t *tmr, uint32_t pwm)
         MXC_TMR_RevB_ClearFlags(tmr); // Clear flags so we can catch the next one
         while (!MXC_TMR_RevB_GetFlags(tmr)) {} // Wait for next PWM transition
         MXC_TMR_RevB_Stop(tmr); // Pause timer
-        MXC_TMR_RevB_SetCount(tmr, 0); // Reset the count
+        MXC_TMR_RevB_SetCount(tmr, 1); // Reset the count
```
Jake-Carter, line 281: What is the desired behavior for PWM == 0?  Do we still want a very small signal in this case?

For https://github.com/Analog-Devices-MSDK/msdk/pull/818  I assumed the desired behavior for a PWM of 0 is no signal (0% duty cycle = no transitions), so I may have been mistaken.

Jake-Carter (approved):
Please confirm the desired behavior of the signal PWM == 0.  Otherwise, looks good.

Libraries/PeriphDrivers/Source/TMR/tmr_revb.c
hfakkiz: By assuming polarity is 1 as defined below picture.
pwm_polarity

Desired behavior for PWM == 0 is continues inactive signal.
pwm_behavior

However, if PWM is set to 0, the output signal is observed like following picture. Because of cnt can not match with pwm value.
pwm_behavior_wrong

If program changes 0 pwm values to 1 like this code does `tmr->pwm = pwm ? pwm : 1;` then output signal act like following picture.
pwm_1

I couldn't find a way to fix this small pulse without a workaround because of timer resets cnt register to 0x1 instead of 0x0. But at least this solution gives %0 duty cycle output signal.

Jake-Carter: > Desired behavior for PWM == 0 is continues inactive signal.

For polarity == 1 the _inactive_ signal is logic high.  So I think that pwm = 0 achieves the desired behavior.  

But in the second image the polarity is 1.  For polarity == 1 the inactive signal is always _high_.  To achieve the desired behavior with pwm == 0, can't the polarity just be set to 0?  Then the inactive signal will always be low, while still achieving the desired result for `pwm=period/2` and `pwm=period` tests.

My concern with overwriting to pwm = 1 is the little pulses in the waveform, so it's a non-zero duty cycle.

Jake-Carter (changes requested):
https://github.com/Analog-Devices-MSDK/msdk/pull/834/files#r1428583186

Libraries/PeriphDrivers/Source/TMR/tmr_revb.c
ozersa: > What is the desired behavior for PWM == 0? Do we still want a very small signal in this case?
> 
> For #818 I assumed the desired behavior for a PWM of 0 is no signal (0% duty cycle = no transitions), so I may have been mistaken.

Desired signal for pwm==0 is no signal, we handled this case by upper layer before calling sdk driver, see below links.
So that, @hfakkiz  will double check pwm=pwm?pwm:1 and MXC_TMR_RevB_SetCount(tmr, 1), if this changes are not necessary for our case, please remove them.

https://github.com/adi-innersource/zephyr/blob/MAX32xxx-Driver-PWM/drivers/pwm/pwm_max32.c
image

hfakkiz: Removed related commit.

Jake-Carter (approved):

Jake-Carter:
Thanks @ozersa and @hfakkiz

PR #845, opened 2023-12-22
feat(PeriphDrivers): Add API to lock Debug Access Port 

Jacob-Scheiffler:
Adding a new function which will PERMANENTLY lock the debug access port. This was requested by Jim, otherwise I think it would be unwise to expose this to customers.


EricB-ADI:
Just a recommendation. If the code is compiled in debug mode, it should probably just return.

BrentK-ADI:
Just came across this looking at open PRs.  A recommendation as an end user: Add the word Permanent to the function name.  Two reasons:

1. Developers tend to look at function names and implement and only read the documentation later when stuff doesn't work
2. Some devices have both temporary and permanent locking features

The MAX32672 for example has the ability to both lock the DAP with unlock capabilities, and permanently lock it.
It might be nice to expose the "temporary" lock capabilities of devices the MXC_SYS_LockDAP & MXC_SYS_UnlockDAP calls. And explicitly call out Permanent in this implementation.

Jake-Carter:
> Just came across this looking at open PRs. A recommendation as an end user: Add the word Permanent to the function name. Two reasons:
> 
> 1. Developers tend to look at function names and implement and only read the documentation later when stuff doesn't work
> 2. Some devices have both temporary and permanent locking features
> 
> The MAX32672 for example has the ability to both lock the DAP with unlock capabilities, and permanently lock it. It might be nice to expose the "temporary" lock capabilities of devices the MXC_SYS_LockDAP & MXC_SYS_UnlockDAP calls. And explicitly call out Permanent in this implementation.

https://github.com/Analog-Devices-MSDK/msdk/pull/845/commits/31dc086dbd05a0ee9217fb007cdbcb5749f02985

Will take a stab at the MAX32672 in a future PR

EricB-ADI (approved):

Issue #762, opened 2023-10-10
The Max78000 FTHR, while running the kws_20_demo with a 3.5mm microphone input, is experiencing issues


ChongHao-Qiu:
I am working on developing an IOT device based on Max7800 FTHR. However, I faced an error when I tried to use the 3.5mm audio input with my own microphone on the FTHR board running kws20_demo. I followed the readme file in the “MAX78000/CNN/kws20_demo/” project and enabled **ENABLE_CODEC_MIC** in the project.mk file. But I got an error in the image below when executing the max9867_init(mc_i2c_regs_t *i2c_inst, int mclk, int controller) function.
![image](https://github.com/Analog-Devices-MSDK/msdk/assets/55849495/14fb1922-b539-4e04-985d-afa5b4abd721)
![image](https://github.com/Analog-Devices-MSDK/msdk/assets/55849495/db977902-3f73-4c98-882a-2ea55fa9a663)

There seems to be something wrong with the **reg_read** function in **_max9867.c_** and the control mode of the I2C1.
Could you help me check this problem? I want to use my own microphone in the kws20_demo.

Jake-Carter:
Hi @ChongHao-Qiu, thanks for reporting.  We will re-test the project to see if we can replicate the issue

Jake-Carter:
We were able to replicate the issue and are investigating the codec driver code

Jake-Carter:
Sorry for the delay @ChongHao-Qiu, 

The following diff fixes the codec drivers.  It turns out the I2C was never initialized...

```diff
diff --git a/Libraries/MiscDrivers/CODEC/max9867.c b/Libraries/MiscDrivers/CODEC/max9867.c
index d895fa0c81..173704495a 100644
--- a/Libraries/MiscDrivers/CODEC/max9867.c
+++ b/Libraries/MiscDrivers/CODEC/max9867.c
@@ -377,6 +377,12 @@ int max9867_init(mxc_i2c_regs_t *i2c_inst, int mclk, int controller)
     if (!i2c_inst)
         return E_NULL_PTR;

+    if ((err = MXC_I2C_Init(i2c_inst, 1, MAX9867_ADDR)) != E_NO_ERROR)
+        return err;
+
+    if ((err = MXC_I2C_SetFrequency(i2c_inst, 100000)) != E_NO_ERROR)
+        return err;
+
     /* Static I2C request values */
     i2c_req.i2c = i2c_inst;
[...]

PR #847, opened 2023-12-28
fix(PeriphDrivers, Examples): MSDK-1232: Update I2S driver for confused usage of wsize and bits_word register fields

ahmetalincak:
### Description

Fix I2S Peripheral Driver for mixed usage of sample_size and bits_word registers. As described in the related Jira tickets, there was a confused usage for the word size and bits_word.

image

image

I updated all I2S examples to use new changes and made them buildable. I am going to open a new PR for examples after my changes are approved.


ahmetalincak:
Also fixes the #762 .

Libraries/PeriphDrivers/Include/MAX32655/i2s.h
```diff
@@ -91,11 +87,12 @@ typedef enum {
...
     mxc_i2s_ch_mode_t channelMode;
     mxc_i2s_stereo_t stereoMode;
-    mxc_i2s_wsize_t wordSize;
     mxc_i2s_justify_t justify;
     mxc_i2s_bitorder_t bitOrder;
     mxc_i2s_polarity_t wsPolarity;
-    mxc_i2s_samplesize_t sampleSize;
+    mxc_i2s_bitsword_t bitsWord;
+    mxc_i2s_wsize_t wordSize;
+    uint8_t sampleSize;
```
sihyung-maxim, line 95: There's the issue of preserving struct member names because current projects using I2S will be using the `mxc_i2s_samplesize_t sampleSize` member and the `mxc_i2s_samplesize_t` enum values instead of how it should be used for `uint8_t sampleSize` and `mxc_i2s_bitwords_t bitsWord`. This adds some complication... Name changes are always difficult to handle as we want to make things clear while also preventing any projects from breaking since these names have been released and in use for quite some time now.

We might need a discussion with @lorne-maxim and @Jake-Carter to figure out what the best approach is for this issue.

sihyung-maxim (commented):
Name change complications... will need to discuss with the team.

Examples/MAX78000/CNN/kws20_demo-riscv/main_riscv.c
```diff
@@ -475,8 +475,8 @@ void I2SInit()
     /* Initialize I2S RX buffer */
     memset(i2s_rx_buffer, 0, sizeof(i2s_rx_buffer));
     /* Configure I2S interface parameters */
-    req.wordSize = MXC_I2S_DATASIZE_WORD;
-    req.sampleSize = MXC_I2S_SAMPLESIZE_THIRTYTWO;
+    req.wordSize = MXC_I2S_WSIZE_WORD;
+    req.bitsWord = MXC_I2S_BITSWORD_THIRTYTWO;
```
Jake-Carter, line 479: @ahmetalincak do we need to initialize `req.SampleSize` here as well?  If we leave it unset I think it will be randomly initialized to any value

Libraries/PeriphDrivers/Include/MAX32655/i2s.h
Jake-Carter: Since `bitsWord` is completely new, I think `sampleSize` is the only issue (after https://github.com/Analog-Devices-MSDK/msdk/pull/847#discussion_r1445237701).

I think this would work:

```C
typedef enum {
    MXC_I2S_SAMPLESIZE_EIGHT = 8,
    MXC_I2S_SAMPLESIZE_SIXTEEN = 16,
    MXC_I2S_SAMPLESIZE_TWENTY = 20,
    MXC_I2S_SAMPLESIZE_TWENTYFOUR = 24,
    MXC_I2S_SAMPLESIZE_THIRTYTWO = 32
} mxc_i2s_samplesize_t __attribute__((deprecated)); // leave samplesize_t defined, but change enum values

typedef enum {
    MXC_I2S_BITSWORD_EIGHT,
    MXC_I2S_BITSWORD_SIXTEEN,
    MXC_I2S_BITSWORD_TWENTY,
    MXC_I2S_BITSWORD_TWENTYFOUR,
    MXC_I2S_BITSWORD_THIRTYTWO,
} mxc_i2s_bitsword_t;
```

So even if application code writes...

```C
req.bitsWord = MXC_I2S_BITSWORD_TWENTYFOUR;
req.sampleSize = MXC_I2S_SAMPLESIZE_EIGHT;
// ...
```
... sampleSize still functionally works as expected with the type change in the new struct definition.

The code actually builds cleanly without any warnings at all.  The only way you can get the deprecation to print is if you write this, which no one will actualy do...  so I don't know if the deprecation attribute is even worth it.

```C
mxc_i2s_samplesize_t test = MXC_I2S_SAMPLESIZE_EIGHT;
```

Libraries/PeriphDrivers/Include/MAX32655/i2s.h
```diff
@@ -64,20 +64,16 @@ typedef enum { MXC_I2S_MSB_JUSTIFY, MXC_I2S_LSB_JUSTIFY } mxc_i2s_justify_t;
...
-    MXC_I2S_SAMPLESIZE_TWENTY,
-    MXC_I2S_SAMPLESIZE_TWENTYFOUR,
-    MXC_I2S_SAMPLESIZE_THIRTYTWO,
-} mxc_i2s_samplesize_t;
+    MXC_I2S_BITSWORD_EIGHT,
+    MXC_I2S_BITSWORD_SIXTEEN,
+    MXC_I2S_BITSWORD_TWENTY,
+    MXC_I2S_BITSWORD_TWENTYFOUR,
+    MXC_I2S_BITSWORD_THIRTYTWO,
+} mxc_i2s_bitsword_t;
```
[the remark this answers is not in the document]
Jake-Carter: https://github.com/Analog-Devices-MSDK/msdk/pull/847#discussion_r1451261706

Jake-Carter (commented):
Thanks @ahmetalincak, this is a great improvement.

@BrentK-ADI I believe the I2S application you were working on last year would be a good case study to check these changes against.  Can you run a build of your application against this branch and give us your perspective?  I wonder how significant the build errors/update requirements are.  Ideally, it's just a few minor updates to your request structs.  Your feedback would be really valuable considering the driver issues you worked around.

Libraries/PeriphDrivers/Include/MAX32655/i2s.h
```diff
@@ -64,20 +64,16 @@ typedef enum { MXC_I2S_MSB_JUSTIFY, MXC_I2S_LSB_JUSTIFY } mxc_i2s_justify_t;
 typedef enum { MXC_I2S_CLKSRC_ERFO, MXC_I2S_CLKSRC_EXT } mxc_i2s_clksrc_t;
 
 /** @brief  I2S transaction word size */
-typedef enum {
-    MXC_I2S_DATASIZE_BYTE,
-    MXC_I2S_DATASIZE_HALFWORD,
-    MXC_I2S_DATASIZE_WORD
-} mxc_i2s_wsize_t;
+typedef enum { MXC_I2S_WSIZE_BYTE, MXC_I2S_WSIZE_HALFWORD, MXC_I2S_WSIZE_WORD } mxc_i2s_wsize_t;
```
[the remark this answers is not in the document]
ahmetalincak: This change is just for following the MSDK convention - I just thought I2S driver needs a full renewal. I prefer not to apply above suggestion as it adds more complexity, we can keep as it is if you think this would cause compatibility problems. Please let me know.

Examples/MAX78000/CNN/kws20_demo-riscv/main_riscv.c
ahmetalincak: Good catch. Will resolve in the next commit.

Libraries/PeriphDrivers/Include/MAX32655/i2s.h
ahmetalincak: Sounds applicable to me. I prefer not to include deprecated stuff in the API. IMHO 'crash and build it clean' or 'do not change at all' would be a good practice for the APIs. Surely, it depends on the popularity of the API to take this radical decision.

Libraries/PeriphDrivers/Include/MAX32655/i2s.h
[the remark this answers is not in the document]
Jake-Carter: @ahmetalincak  I think this is a good hybrid approach.  It preserves the legacy compatibility but also allows for setting the sample size to integer values if needed.  I like this better than deprecating the old struct completely.

Libraries/PeriphDrivers/Include/MAX32655/i2s.h
Jake-Carter: I will resolve this thread in favor of https://github.com/Analog-Devices-MSDK/msdk/pull/847#discussion_r1452416765

Libraries/PeriphDrivers/Include/MAX32655/i2s.h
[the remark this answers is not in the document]
Jake-Carter: We've implemented these enum "pass-throughs" to avoid compilation errors in some other drivers.  So there is a precedent for doing this.  We must consider projects that are even 3-4 years old.  If we can avoid breaking them, then we must implement it.

I'm in favor of your change to match the enums to the typdef name.  It's annoying when I'm writing code but the enum does not match what I expect.  This also matches the field name from the UG.

I also vote for the additional pass-throughs to preserve backwards compatibility.  We should add comments/doxygen to them, though.

```C
/** @brief I2S transaction word size.  
 * 
 * Set this field to the desired width for data writes and reads from the FIFO.  
 * This should be greater than or equal to the number of bits per word.  */
typedef enum { 
    MXC_I2S_WSIZE_BYTE, ///< Set 8-bit FIFO transactions
    MXC_I2S_WSIZE_HALFWORD, ///< Set 16-bit FIFO transactions
    MXC_I2S_WSIZE_WORD ///< Set 32-bit FIFO transactions

    MXC_I2S_DATASIZE_BYTE = MXC_I2S_WSIZE_BYTE, ///< Legacy name.  Use MXC_I2S_WSIZE_BYTE instead.
    MXC_I2S_DATASIZE_HALFWORD = MXC_I2S_WSIZE_HALFWORD, ///< Legacy name.  Use MXC_I2S_WSIZE_HALFWORD instead.
    MXC_I2S_DATASIZE_WORD = MXC_I2S_WSIZE_WORD, ///< Legacy name.  Use MXC_I2S_WSIZE_WORD instead.
} mxc_i2s_wsize_t;
```

Jake-Carter (approved):

Examples/MAX32655/I2S/main.c
```diff
@@ -74,8 +74,10 @@ int main()
     memset(i2s_rx_buffer, 0, sizeof(i2s_rx_buffer));
 
     /* Configure I2S interface parameters */
-    req.wordSize = MXC_I2S_DATASIZE_WORD;
+    req.wordSize = MXC_I2S_WSIZE_WORD;
     req.sampleSize = MXC_I2S_SAMPLESIZE_THIRTYTWO;
+    req.bitsWord = 32 - 1;
```
sihyung-maxim, line 79: The I2S drivers can handle the `- 1` for the user because it's a register-specific rule. The user should just focus on what `bitsWord` value they need.

You can reference what we do for setting the [DMA burst size](https://github.com/Analog-Devices-MSDK/msdk/blob/main/Libraries/PeriphDrivers/Source/DMA/dma_reva.c#L209). The user passes in the burst size they want, and the DMA drivers properly write the decremented value to the register.

sihyung-maxim (changes requested):

sihyung-maxim:
Everything else looks good, by the way.

ahmetalincak:
Hi @Jake-Carter, thanks for the latest change!

Jake-Carter:
> Hi @Jake-Carter, thanks for the latest change!

No problem!

Examples/MAX32655/I2S/main.c
Jake-Carter: Fixed in https://github.com/Analog-Devices-MSDK/msdk/pull/847/commits/27c49436b024eb2e392c597335da0d02698d3aca

selimgullulu:
Hi, I'm having problems using the example on the MAX78000FTHR using codec and external mic.
Please see it [here](https://github.com/analogdevicesinc/msdk/issues/1266)

PR #852, opened 2024-01-02
fix(MiscDrivers): MSDK-1240: Clear out of bounds warnings seen with GCC v13.2

ahmetalincak:
### Description

GCC 13.2 warning clearance for tft_ssd2119.c

In the latest compiler version, the array bounds warnings appear. This was an internal issue and I tried building MAX78000 kws20_demo example and saw warnings myself. 

The changes are in:
1. MXC_TFT_Init function: Optional and tested for both GCC Versions (10.3 and 13.2) on MAX78000 kws20_demo.
2. MXC_TFT_Printf function: This one works incorrect when using GCC v13.2 - so should be merged. Tested with MAX32662 Demo Example.

### DIFF
--- a/Libraries/MiscDrivers/Display/tft_ssd2119.c
+++ b/Libraries/MiscDrivers/Display/tft_ssd2119.c
@@ -118,8 +118,8 @@ typedef struct {
 
 #pragma pack()
 
-extern Header_images_t _bin_start_; // binary start address, defined in linker file
-extern unsigned int _bin_end_; // binary end address, defined in linker file
+extern unsigned char _bin_start_[]; // binary start address, defined in linker file
+extern unsigned char _bin_end_; // binary end address, defined in linker file
 static unsigned char *images_start_addr = NULL;
 static Header_images_t images_header;
 
@@ -927,10 +927,10 @@ int MXC_TFT_Init(void)
     memset(&images_header, 0, sizeof(Header_images_t));
 
     // Is there any image data to work with?
-    if ((unsigned int *)&_bin_start_ != (unsigned int *)&_bin_end_) {
-        images_start_addr = (unsigned char *)&_bin_start_;
+    if (_bin_start_ != &_bin_end_) {
+        images_start_addr = _bin_start_;
         // set header
-        images_header = *((Header_images_t *)&_bin_start_);
+        memcpy(&images_header, images_start_addr, sizeof(Header_images_t));
     }
 
     /*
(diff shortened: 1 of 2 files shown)


Libraries/MiscDrivers/Display/tft_ssd2119.c
```diff
@@ -926,10 +927,10 @@ int MXC_TFT_Init(void)
     memset(&images_header, 0, sizeof(Header_images_t));
 
     // Is there any image data to work with?
-    if (_bin_start_ != _bin_end_) {
+    if ((unsigned int *)&_bin_start_ != (unsigned int *)&_bin_end_) {
```
rotx-maxim, line 930: Is this cast necessary, or should this reference the first field (unsigned int offset2info_palatte) of Header_images_t instead?

Libraries/MiscDrivers/Display/tft_ssd2119.c
```diff
@@ -926,10 +927,10 @@ int MXC_TFT_Init(void)
...
         images_start_addr = (unsigned char *)&_bin_start_;
         // set header
-        memcpy(&images_header, images_start_addr, sizeof(Header_images_t));
+        images_header = *((Header_images_t *)&_bin_start_);
```
rotx-maxim, line 933: Why is this cast needed?

Libraries/MiscDrivers/Display/tft_ssd2119.c
ahmetalincak: Casting is for readability purposes only. We're checking if any data inserted in this specific block.

Libraries/MiscDrivers/Display/tft_ssd2119.c
ahmetalincak: GCC 13.2 wants to be informed about the memory address is enough for the related memcopy. We're giving a struct type pointer which we claim it will not overflow with the memory copying. Actually *(&_bin_start_) is enough but I added (Header_images_t *) casting for readability.

Libraries/MiscDrivers/Display/tft_ssd2119.c
rotx-maxim: These casts are the opposite of "readable" and like any cast in C, carry the risk of mis-aligned accesses and access to unallocated memories.

Libraries/MiscDrivers/Display/tft_ssd2119.c
rotx-maxim: This does not add to readability, it reduces readability and adds the possibility of mis-aligned access, and access to unallocated memories, and incompatible data structures.
This should be
images_header = _bin_start_;

rotx-maxim (changes requested):
Improve casts

rotx-maxim (approved):

Jake-Carter (approved):
Thanks @ahmetalincak

PR #867, opened 2024-01-12
fix(CMSIS): MSDK-1249: Add OVR Correction to SystemCoreClock Calculation for MAX32662 MAX32672 MAX32675

ahmetalincak:
### Description

MAX32662, MAX32672 and MAX32675 has the OVR bit to operate CPU voltage with different modes and this affects also the IPO clock. We need to consider this when calculating the system core clock.

PS. This has already done for MAX32660 in [here](https://github.com/Analog-Devices-MSDK/msdk/blob/b71d6f576b70670144f35bf898ddd7ef4e92c53a/Libraries/CMSIS/Device/Maxim/MAX32660/Source/system_max32660.c#L65)


Jake-Carter (approved):
Thanks @ahmetalincak looks good!

PR #868, opened 2024-01-12
fix(PeriphDrivers): Add I2C RX FIFO status check to receive more than 256-byte of data for I2C synchronous transactions

ahmetalincak:
### Description

@alperenguclu faced a problem while trying to receive more than 256-byte data from I2C channel and we found that `MXC_I2C_RevA_MasterTransaction` has a bug in rx part. His team tested it with MAX32690.


alperenguclu (approved):

sihyung-maxim (approved):

Jake-Carter (approved):

Jake-Carter (approved):
Thanks @alperenguclu - it looks like this works around some FIFO issue at the end of each 256-byte chunk, is that right?

Any details you can share on the hardware issue may be helpful at some point in the future

PR #869, opened 2024-01-12
fix(Examples): Fix MAX78002 TFT_Demo Example for TFT=ADAFRUIT

selimgullulu:
TFT_Demo Example for MAX78002 with Adafruit Featherwing TFT Display was not working properly.

-main.c is changed in order to initiate the TFT Display, then the Touchscreen.

After the change, the example works fine:
IMG_8172
IMG_8173
IMG_8174


sihyung-maxim (approved):

Jake-Carter (approved):
Thanks @selimgullulu

PR #871, opened 2024-01-12
 fix(BLE): Update LIBPHY ME17B

EricB-ADI:
### Description
Update LIBPHY ME17B


EdwinFairchild:
whats the update? does it have that mic fix you did the other day?

PR #878, opened 2024-01-16
fix(PeriphDrivers, Examples): MSDK-1232: Update I2S driver for confused usage of wsize and bits_word register fields - PR after discussions

ahmetalincak:
### Description

After discussions and careful considerations, I decided to create another PR for #MSDK-1232. This description is a clone of #847.

Fix I2S Peripheral Driver for mixed usage of sample_size and bits_word registers. As described in the related Jira tickets, there was a confused usage for the word size and bits_word.

image

image

I updated all I2S examples to use new changes and made them buildable. I am going to open a new PR for examples after my changes are approved.

### DIFF
--- a/Examples/MAX32655/I2S/main.c
+++ b/Examples/MAX32655/I2S/main.c
@@ -74,7 +74,7 @@ int main()
     memset(i2s_rx_buffer, 0, sizeof(i2s_rx_buffer));
 
     /* Configure I2S interface parameters */
-    req.wordSize = MXC_I2S_DATASIZE_WORD;
+    req.wordSize = MXC_I2S_WSIZE_WORD;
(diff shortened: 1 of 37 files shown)


Jake-Carter:
Thanks @ahmetalincak, will take another look this week

Jake-Carter (changes requested):
Hi @ahmetalincak, I think I prefer the solution in https://github.com/Analog-Devices-MSDK/msdk/pull/847#discussion_r1452416765

We have gone with extended config APIs in some cases in the past, but in my opinion they complicate the drivers too much.

@sihyung-maxim what do you think?

ahmetalincak:
Falling back to #847

PR #880, opened 2024-01-16
fix(PeriphDrivers): MSDK-1245: Re-order VSSEL selection to avoid possible damages on pads

ahmetalincak:
### Description

Voltage Rail selection should be made before padding configuration. A hesitation came from Clayton:

> For the MXC_GPIO_Config function, the last thing done is to select the voltage rail (gpio_me18.c). This creates a situation where damage to the part can occur. The GPIO on parts with multiple rails defaults to VDDIO (1.8V) with pullup/pulldown disabled. If a customer has 3.0V on a GPIO pin (say I2C pullups), and the code turns on the pullup before selecting the VDDIOH rail, the 3.0V back powers VDDIO, potentially causing damage to the pads.


sihyung-maxim (approved):

Jake-Carter (approved):
This makes sense.  Thanks @cwarer1 and @ahmetalincak

Issue #949, opened 2024-03-08
MAX32690/ME18 Alternate I2C


BrentK-ADI:
The ME18 periph drivers does not include support for alternate pins for I2C1 (P2.18/2.17).  Without being able to specify the alternate pins during init, the MXC_I2C_Recover call which is issued during the initialization process fails, causing the initialization itself to fail.

Current work around is modifying pins_me18.c to specify the alternate pin configuration.

Other I2C instances and peripherals should be checked for similar problems.

Jake-Carter:
@BrentK-ADI I think this might be a typo in the [datasheet](https://www.analog.com/media/en/technical-documentation/data-sheets/max32690.pdf) (?).  I2C1A is listed on AF1 as P0.11/P0.12 _and_ P2.17/P2.18 for the 140 WLP...

BrentK-ADI:
@Jake-Carter I agree there's a typo/confusing naming in the datasheet.  The same applies for P0.30/0.31 & P2.7/P2.8 (I2C0A) and P0.12/0.13 & P1.7/1.8 (I2C2C).    However, datasheet aside, these are valid alternate pins for the I2C peripherals.  I verified on the APARD board which has the 140WLP.

Also note the [errata](https://www.analog.com/media/en/technical-documentation/data-sheets/max32690_a2_errata_rev2.pdf) regarding these pins for the RevA2 silicon:
![image](https://github.com/analogdevicesinc/msdk/assets/129296493/be98bff4-d3d5-4674-9d67-49fe3004b31c)

Jake-Carter:
I see, thanks.  I wasn't aware of the errata.

I think I'll update the definitions to enable both sets of pins simultaneously

PR #950, opened 2024-03-11
feat(Boards): AD-APARD32690-SL Board Support

BrentK-ADI:
### Description

Initial commit of the board support for the AD-APARD32690-SL board for the MAX32690.  Board files and updates to GPIO example project. [AD-APARD32690](https://wiki.analog.com/resources/eval/user-guides/ad-apard32690-sl)


BrentK-ADI:
Commits so far include the actual board files and updates to the GPIO example, which looked to be the only example impacted by the different pin assignments.

Provide guidance on additional changes needed to complete the PR.

Jake-Carter (changes requested):
Thanks @BrentK-ADI BSP looks good!  Below are some remaining items:

- [ ] Looks like the TFT demo also needs a little patching.  From the auto-tester:

```shell
MAX32690 (APARD): TFT_Demo failed.
Build command: make -r -j 8 TARGET=MAX32690 MAXIM_PATH=/home/runner/work/msdk/msdk BOARD=APARD FORCE_COLOR=1
Errors:
----------------------------------------
main.c:38:10: fatal error: tft_st7735.h: No such file or directory
   38 | #include "tft_st7735.h"
      |          ^~~~~~~~~~~~~~
compilation terminated.
make: *** 
[/home/runner/work/msdk/msdk/Libraries/CMSIS/Device/Maxim/GCC/gcc.mk:385: 
/home/runner/work/msdk/msdk/Examples/MAX32690/TFT_Demo/build/main.o] Error 1
make: *** Waiting for unfinished jobs....
----------------------------------------
```

FYI if you don't want to support a specific example (i.e. the board doesn't have a TFT), you can throw a `ERR_NOTSUPPORTED` from project.mk.  The build tester will log it as a known condition and pass it, and this will block mistaken users as well.

```Makefile
# project.mk
ifeq "$(BOARD)" "APARD"
$(error ERR_NOTSUPPORTED: This project is not supported for the APARD32690 board)
endif
```

- [ ] Add part under MAX32690 [support parts](https://github.com/analogdevicesinc/msdk/blob/main/USERGUIDE.md#supported-parts) in UG
- [ ] Add part to [BSP table](https://github.com/analogdevicesinc/msdk/blob/main/USERGUIDE.md#bsp-table) in UG
- [ ] clang-format (see [contributing](https://analogdevicesinc.github.io/msdk//CONTRIBUTING/#style-guide))

BrentK-ADI:
@Jake-Carter Thanks for the tip on the ERR_NOTSUPPORTED.  This board doesn't have a display so might make sense to skip that example. 

I may have missed it, but didn't see any BSPs that include HyperRAM in them.  This board has 2 banks of HyperRAM attached to CS0 and CS1 respectively.  Should this be added similar to the External Flash, or leave it separate as just an example application?

Also, not for this commit, but future work, the board has a ADIN1110 MACPHY and MAXQ1065 Crypto Co-Processor.  At some point it'd be nice to tie these in, although we'll have to navigate the licensing issues as those SDKs fall under different licenses than the MSDK.

sihyung-maxim:
> @Jake-Carter Thanks for the tip on the ERR_NOTSUPPORTED. This board doesn't have a display so might make sense to skip that example.
> 
> I may have missed it, but didn't see any BSPs that include HyperRAM in them. This board has 2 banks of HyperRAM attached to CS0 and CS1 respectively. Should this be added similar to the External Flash, or leave it separate as just an example application?
> 
> Also, not for this commit, but future work, the board has a ADIN1110 MACPHY and MAXQ1065 Crypto Co-Processor. At some point it'd be nice to tie these in, although we'll have to navigate the licensing issues as those SDKs fall under different licenses than the MSDK.

There exists an example using the MAX32690's HyperBus (HPB) Memory Controller peripheral with the HyperRAM.

https://github.com/analogdevicesinc/msdk/tree/main/Examples/MAX32690/HBMC

BrentK-ADI:
The only examples that look to be problematic are anything based on I2C (EEPROM Emulator, I2C, Temp_Monitor).  The APARD board uses the alternate function I2C pins which are not automatically supported by the MSDK. See #949.  We can hold off on this PR until that is resolved, or come back to it later.

Jake-Carter (approved):
@BrentK-ADI looks good!  I opened #958 to fix those pin definitions.

> I may have missed it, but didn't see any BSPs that include HyperRAM in them. This board has 2 banks of HyperRAM attached to CS0 and CS1 respectively. Should this be added similar to the External Flash, or leave it separate as just an example application?

The more you can include in the BSP to match the on-board hardware the better.  I would suggest adding it similar to the external flash

BrentK-ADI:
Rebased against the latest main which includes the I2C fixes.  Updated I2C related examples.  Needed to error out on two of them:

- I2C - Due to the errata and the use of the alternate pins on the board, 2 I2C buses arent available to perform the loopback example
- Temp_Monitor - No provisions on the board for 1.8V I2C

Libraries/Boards/MAX32690/APARD/examples.txt
```diff
@@ -0,0 +1,71 @@
...
+Dual_Core_Sync/Dual_core_sync_arm
+Dual_Core_Sync/Dual_core_sync_riscv
+EEPROM_Emulator
+Flash
+Flash_CLI
+FreeRTOSDemo
+GPIO
+HBMC
+Hello_World
+I2C
```
Jake-Carter, line 44: Remove unsupported examples from examples.txt as well (this is what Eclipse uses to populate its dropdown)

Libraries/Boards/MAX32690/APARD/Include/board.h
```diff
@@ -0,0 +1,125 @@
...
+#ifdef LED_ON
+#undef LED_ON
+#endif
+#define LED_ON 1 /// Override active state of LEDs
+
+#ifndef EXT_FLASH_BAUD
+#define EXT_FLASH_BAUD 4000000
+#endif
+
+#define APARD /// Used in examples to control program flow
```
Jake-Carter, line 73: Just noticed this (minor quibble) - `BOARD_APARD` is defined on line 37.  Existing conventions prefers using `BOARD_xxx` for program flow

Libraries/Boards/MAX32690/APARD/examples.txt
BrentK-ADI: Removed I2C and Temp_Monitor in latest push

Libraries/Boards/MAX32690/APARD/Include/board.h
BrentK-ADI: Updated as requested.

Jake-Carter (approved):

PR #951, opened 2024-03-14
fix(BLE): PalTimer Calibration

BrentK-ADI:
Updated the code for calibrating the PalTimer against the 32MHz clock.  Original implementation only added adjustment on 100ms (PAL_TMR_CALIB_TIME) boundaries.  In cases where the timer time was in between 100ms intervals, the timer could be severely early or late depending on how large usecDiff is.
``
The math could be accomplished by simply running `(expUsec * palTimerCb.usecDiff) / PAL_TMR_CALIB_TIME;` however there is not enough objective evidence for the max values of expUsec and usecDiff to show the multiplication won't overflow 32-bits.  The math could be done in 64-bits, however the modulo method was used to keep this within a 32-bit scope.

### DIFF
--- a/Libraries/Cordio/platform/targets/maxim/max32665/sources/pal_timer.c
+++ b/Libraries/Cordio/platform/targets/maxim/max32665/sources/pal_timer.c
@@ -317,7 +317,7 @@ void PalTimerDeInit(void)
 /*************************************************************************************************/
 /*!
  *  \brief      Return scheduler timer state.
- 
+
  */
 /*************************************************************************************************/
 void PalTimerExecCallback(void)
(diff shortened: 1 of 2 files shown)


EricB-ADI:
Yeah this is good, I am wondering why you cast the uint32_t to an int. It seems intentional, but not obvious as to why?

BrentK-ADI:
usecDiff can be negative, and the with the implicit conversion between the signed and unsigned types, I was getting incorrect values.    

Looking at it though, int should probably be int32_t for consistency.  Let me push up a change using the stdint type instead.

BrentK-ADI:
Pushed up the wrong branch and made a big mess. 6f5ab8b should be good now.

EricB-ADI (approved):

PR #956, opened 2024-03-18
fix(Examples,PeriphDrivers): Update TMR Clock Source names to be consistent with the User Guides

sihyung-maxim:
### Description

The microcontrollers with the TMR_RevB IP Core allow users to select different clock sources for each timer instance. In the MSDK, this is represented using the `mxc_tmr_clock_t` enum. However, the options in the `mxc_tmr_clock_t` enum were named after frequencies of the clock sources, not their individually-referred names.
```
typedef enum {
    MXC_TMR_APB_CLK = 0,
    MXC_TMR_EXT_CLK,
    MXC_TMR_32M_CLK,
    MXC_TMR_8M_CLK,
    MXC_TMR_32K_CLK,
    MXC_TMR_8K_CLK
} mxc_tmr_clock_t;
```

This can cause confusion as the named frequencies doesn't always match the actual frequency of the clock source. For example, the `MXC_TMR_32M_CLK` represents the ERFO clock which could range from 16-32MHz. We have parts in the MSDK setting the ERFO to 16MHz, 25MHz, and 32MHz. The `MXC_TMR_8K_CLK` represents the INRO clock, and this oscillator's frequency is highly variable. Some parts also have the INRO set to 80KHz.

This PR updates the TMR enum names (while leaving the old names in) to make it more readable and consistent with the user guides for the affected parts. Also, I scrubbed through the TMR RevB drivers to ensure the clock source selections are correct and only had what the part actually supports.


Jake-Carter (approved):
Good change, just need to clang/lint

PR #975, opened 2024-04-02
feat(BLE): WSF-NVM Storage Full Error Handling

EricB-ADI:
### Description
- WsfNvm no longer asserts on storage full. Returns a FALSE and leaves it up to user what to do next.
- Added API to de-fragment NVM. Previously once marked as unused all NVM needed to be wiped to use the space again.

### DIFF
--- a/Examples/MAX32655/Bluetooth/BLE_fit_FreeRTOS/project.mk
+++ b/Examples/MAX32655/Bluetooth/BLE_fit_FreeRTOS/project.mk
@@ -23,9 +23,3 @@ INIT_CENTRAL = 0
 # Set to 1 to enable serial port trace messages
 # Set to 2 to enable verbose messages
 TRACE = 1
-
-USE_DUAL_CORE ?= 0
-ifeq "$(USE_DUAL_CORE)" "1"
-BLE_HOST = 1
-BLE_CONTROLLER = 0
-endif
--- a/Examples/MAX32655/Bluetooth/BLE_fit_FreeRTOS/project.mk
+++ b/Examples/MAX32655/Bluetooth/BLE_fit_FreeRTOS/project.mk
@@ -23,3 +23,9 @@ INIT_CENTRAL = 0
 # Set to 1 to enable serial port trace messages
 # Set to 2 to enable verbose messages
 TRACE = 1
+
+USE_DUAL_CORE ?= 0
+ifeq "$(USE_DUAL_CORE)" "1"
+BLE_HOST = 1
+BLE_CONTROLLER = 0
+endif
--- a/Libraries/Cordio/wsf/sources/targets/baremetal/wsf_nvm.c
+++ b/Libraries/Cordio/wsf/sources/targets/baremetal/wsf_nvm.c
@@ -27,7 +27,7 @@
 #include "wsf_assert.h"
 #include "pal_flash.h"
 #include "util/crc32.h"
-
+#include "mxc_device.h"
 /**************************************************************************************************
   Macros
 **************************************************************************************************/
@@ -74,7 +74,14 @@ static struct
 /**************************************************************************************************
   Global Functions
 **************************************************************************************************/
-
+/*************************************************************************************************/
+/*!
+ *  \brief  Check if NVM space is full.
+ */
+/*************************************************************************************************/
+static inline bool_t wsfNvmIsFull(void){
+  return (wsfNvmCb.availAddr - WSF_NVM_START_ADDR) <= wsfNvmCb.totalSize;
+}
 /*************************************************************************************************/
 /*!
  *  \brief  Initialize the WSF NVM.
@@ -252,7 +259,14 @@ bool_t WsfNvmWriteData(uint64_t id, const uint8_t *pData, uint16_t len, WsfNvmCo
   }
 
   WSF_ASSERT(!((id == WSF_NVM_RESERVED_FILECODE) || (id == WSF_NVM_UNUSED_FILECODE)));
-  WSF_ASSERT((wsfNvmCb.availAddr - WSF_NVM_START_ADDR) <= wsfNvmCb.totalSize);
+  
+  if(wsfNvmIsFull()){
+    
+    WSF_TRACE_INFO0("WsfNvm: Failed to write flash! Out of space.");
+    return FALSE;
+  }
+  
+
 
   /* Read first header. */
   PalFlashRead(&header, sizeof(header), storageAddr);
--- a/Libraries/Cordio/wsf/sources/targets/baremetal/wsf_nvm.c
+++ b/Libraries/Cordio/wsf/sources/targets/baremetal/wsf_nvm.c
@@ -21,12 +21,16 @@
  *  limitations under the License.
  */
 /*************************************************************************************************/
+#include <string.h>
+#include <stdint.h>
 
+#include "wsf_cs.h"
 #include "wsf_types.h"
 #include "wsf_nvm.h"
 #include "wsf_assert.h"
 #include "pal_flash.h"
 #include "util/crc32.h"
+#include "wsf_buf.h"
 #include "mxc_device.h"
 /**************************************************************************************************
   Macros
@@ -81,8 +85,18 @@ static struct
  *  \brief  Check if NVM space is full.
  */
 /*************************************************************************************************/
+static inline uint32_t wsfNvmGetRemainingSpace(void){
+  return wsfNvmCb.totalSize - (wsfNvmCb.availAddr - WSF_NVM_START_ADDR);
+}
 static inline bool_t wsfNvmIsFull(void){
-  return !((wsfNvmCb.availAddr - WSF_NVM_START_ADDR) <= wsfNvmCb.totalSize);
+  return wsfNvmGetRemainingSpace() > 0;
+}
+static inline bool_t wsfNvmHaveEnoughSpace(uint32_t lenNeeded){
+  if(lenNeeded > wsfNvmGetRemainingSpace()){
+    return FALSE;
+  }else{
+    return TRUE;
+  }
 }
 /*************************************************************************************************/
 /*!
@@ -267,8 +281,7 @@ bool_t WsfNvmWriteData(uint64_t id, const uint8_t *pData, uint16_t len, WsfNvmCo
 
   WSF_ASSERT(!((id == WSF_NVM_RESERVED_FILECODE) || (id == WSF_NVM_UNUSED_FILECODE)));
   
-  if(wsfNvmIsFull()){
-    
+  if(!wsfNvmHaveEnoughSpace(len)){
     WSF_TRACE_INFO0("WsfNvm: Failed to write flash! Out of space.");
     return FALSE;
   }
@@ -397,6 +410,7 @@ bool_t WsfNvmEraseData(uint64_t id, WsfNvmCompEvent_t compCback)
         header.dataCrc = 0;
         PalFlashWrite(&header, sizeof(header), storageAddr);
         erased = TRUE;
+        break;
       }
     }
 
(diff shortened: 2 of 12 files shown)


Examples/MAX32655/Bluetooth/BLE_fit_FreeRTOS/project.mk
yc-adi, line 1: We'd better test USE_DUAL_CORE = 1 first.

yc-adi (approved):

Libraries/Cordio/wsf/sources/targets/baremetal/wsf_nvm.c
```diff
@@ -402,14 +421,76 @@ bool_t WsfNvmEraseData(uint64_t id, WsfNvmCompEvent_t compCback)
...
+    if (compCback) {
+        compCback(TRUE);
+    }
+}
+
+/*************************************************************************************************/
+/*!
+ *  \brief  Defragment NVM
+ *  \param  copyBuf       Scratch buffer to temporarily copy NVM stored data.
+ *  \param  size          Size of copyBuf.
```
Jake-Carter, line 439: Suggest clarifying the unit ("Size of copyBuf in bytes")

Libraries/Cordio/wsf/sources/targets/baremetal/wsf_nvm.c
```diff
@@ -402,14 +421,76 @@ bool_t WsfNvmEraseData(uint64_t id, WsfNvmCompEvent_t compCback)
...
+        currentOffset += fileSize;
+        PalFlashRead(&header, WSF_NVM_HEADER_SIZE, currentOffset);
+    }
+
+    /*
+    If we defragged anything clear the sector and rewrite flash with defragmented data
+  */
+    if (defragPossible) {
+        WsfNvmEraseDataAll(NULL);
+        PalFlashWrite(copyBuf, copyOffset, wsfNvmCb.availAddr);
```
Jake-Carter, line 487: Might want to error check and trace these last two erase/write operations as well just in case anything goes wrong here

Jake-Carter (changes requested):
Thanks @EricB-ADI, looks good.  Minor requests below

Libraries/Cordio/wsf/sources/targets/baremetal/wsf_nvm.c
EricB-ADI: Both functions return void.

Libraries/Cordio/wsf/sources/targets/baremetal/wsf_nvm.c
```diff
@@ -21,59 +21,88 @@
...
+/*************************************************************************************************/
+/*!
+ *  \brief  Get remaining space in WSF Allcoated NVM.
+ *  \return Bytes left.
+ */
+/*************************************************************************************************/
+int32_t WsfNvmGetRemainingSpace(void)
+{
+    const int32_t bytesLeft = wsfNvmCb.totalSize - (wsfNvmCb.availAddr - WSF_NVM_START_ADDR);
+    WSF_ASSERT(bytesLeft >= 0);
```
sullivanmj, line 93: I tested out the functionality and got caught up here. So would the recommendation be to disable asserts to prevent that, or should this just return false in that case?

Libraries/Cordio/wsf/sources/targets/baremetal/wsf_nvm.c
```diff
@@ -172,63 +195,72 @@ void WsfNvmInit(void)
...
+ *  \brief  Check whether or not the amount of data has space to fit into NVM
+ *
+ *  \param  lenNeeded         Number of bytes wanting to store
+ *
+ *  \return TRUE if NVM operation is successful, FALSE otherwise.
+ */
+/*************************************************************************************************/
+static inline bool_t wsfNvmHaveEnoughSpace(uint32_t lenNeeded)
+{
+    if (lenNeeded > WsfNvmGetRemainingSpace()) {
```
sullivanmj, line 258: This comparison between unsigned `lenNeeded` and the signed return value of `WsfNvmGetRemainingSpace()` can be problematic in some cases. When `WsfNvmGetRemainingSpace()` is negative, this expression evaluates to `FALSE`, which causes the calling code to be unaware that there is not enough space for the write to succeed.

Changing `lenNeeded` to an `int32_t` seems to have the desired effect.

Libraries/Cordio/wsf/sources/targets/freertos/wsf_nvm.c
```diff
@@ -373,3 +423,65 @@ void WsfNvmEraseDataAll(WsfNvmCompEvent_t compCback)
...
+    WsfNvmHeader_t header = { 0 };
+
+    /* Get the first header*/
+    PalFlashRead(&header, WSF_NVM_HEADER_SIZE, currentOffset);
+    while (currentOffset < wsfNvmCb.availAddr && header.id != WSF_NVM_UNUSED_FILECODE) {
+        const uint32_t fileSize = WSF_NVM_FILE_SIZE(header.len);
+
+        /*If the entry is valid copy it into the defragmentation buffer, if its reserved we can defrag it*/
+        if (header.id != WSF_NVM_RESERVED_FILECODE) {
+            PalFlashRead(&copyBuf[copyOffset], fileSize, currentOffset);
```
sullivanmj, line 463: I am getting a HardFault on the `memcpy` call inside this call.

I have defined `PAL_NVM_SIZE` to be 0x2000. I also fixed the bug in [this comment](https://github.com/analogdevicesinc/msdk/issues/798#issuecomment-2038424644).

I filled up the DB, and now when this code executes, `wsfNvmCb.availAddr`'s value is 0x2010, which is past the bounds of the DB. So the `header` actually has invalid data. `header.len` is 0x2000242c, which ultimately causes the HardFault in `memcpy`.

EricB-ADI: Can you see where the address gets set to the out of bounds state? Once again the ```wsfHaveEnoughSpace``` function is intended to block against this condition.  If it isn't, I need to fix that. A workaround that I don't like is just getting the min between the available address and the total size

sullivanmj: It's being set in `WsfNvmInit` on a subsequent boot, after the failed write due to insufficient space.

Libraries/Cordio/wsf/sources/targets/baremetal/wsf_nvm.c
EricB-ADI: As mentioned in the previous comment, I updated the code to return an unsigned value, removing this issue.

Libraries/Cordio/wsf/sources/targets/baremetal/wsf_nvm.c
EricB-ADI: After running this code, restarting and running ```WsfNvmInit``` the available address is correct 

```    WsfNvmInit();

    uint8_t data[1000] = {0};
    bool_t writeOk = TRUE;
    uint32_t id = 1;

    while(writeOk)
    {
        writeOk = WsfNvmWriteData(id, data, sizeof(data), NULL);
        id++;
    }
```

```
(gdb) p wsfNvmCb
$37 = {
  availAddr = 16384,
  sectorSize = 8192,
  totalSize = 16384
}
```

Libraries/Cordio/wsf/sources/targets/freertos/wsf_nvm.c
```diff
@@ -232,7 +287,11 @@ bool_t WsfNvmWriteData(uint64_t id, const uint8_t *pData, uint16_t len, WsfNvmCo
   }
 
   WSF_ASSERT(!((id == WSF_NVM_RESERVED_FILECODE) || (id == WSF_NVM_UNUSED_FILECODE)));
-  WSF_ASSERT((wsfNvmCb.availAddr - WSF_NVM_START_ADDR) <= wsfNvmCb.totalSize);
+
+  if (!wsfNvmHaveEnoughSpace(len)) {
```
sullivanmj, line 291: Should this be checking for `wsfNvmHaveEnoughSpace(len+sizeof(header))`? I am still seeing a HardFault, even with the additional fixes from today (which look great to me). I'm trying to figure out if it's a problem with my test setup or not.

Libraries/Cordio/wsf/sources/targets/freertos/wsf_nvm.c
```diff
@@ -373,3 +432,65 @@ void WsfNvmEraseDataAll(WsfNvmCompEvent_t compCback)
...
+
+        currentOffset += fileSize;
+        PalFlashRead(&header, WSF_NVM_HEADER_SIZE, currentOffset);
+    }
+
+    /*
+    If we defragged anything clear the sector and rewrite flash with defragmented data
+  */
+    if (defragPossible) {
+        WsfNvmEraseDataAll(NULL);
```
sullivanmj, line 486: I am seeing another HardFault inside this call, because inside `WsfNvmEraseDataAll`, `wsfNvmCb.availAddr` is not rounded up to the next nearest multiple of `wsfNvmCb.sectorSize`. So `PalFlashEraseSector` enters an endless loop of calling `MXC_FLC_PageErase`.

Libraries/Cordio/wsf/sources/targets/freertos/wsf_nvm.c
```diff
@@ -63,6 +65,9 @@ typedef struct {
   uint32_t          dataCrc;    /*!< CRC of subsequent data. */
 } WsfNvmHeader_t;
 
+#define WSF_NVM_HEADER_SIZE sizeof(WsfNvmHeader_t)
+#define WSF_NVM_FILE_SIZE(header_len) (WSF_NVM_HEADER_SIZE + (header_len))
```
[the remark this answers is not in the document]
sullivanmj: I forgot to add, I also had to change this line https://github.com/analogdevicesinc/msdk/blob/ac55c78796e717a68e744a3c95a8f61f70a431cd/Libraries/Cordio/wsf/sources/targets/freertos/wsf_nvm.c#L478

to account for that alignment offset as well, in order to fix the HardFault

Libraries/Cordio/wsf/sources/targets/freertos/wsf_nvm.c
EricB-ADI: Yes, i believe that is correct. I updated the macro to deal with alignment.

EricB-ADI: I can get the hardfaults now, and I am working through them.

EricB-ADI:
Hardfaults are gone and non word aligned de fragmentation is fixed

Libraries/Cordio/wsf/sources/targets/freertos/wsf_nvm.c
sullivanmj: I see that the first HardFault is fixed with the latest code that is pushed, but am still getting a hardfault here, if I try to defrag and `wsfNvmCb.availAddr` is not rounded up to the start of the next sector. This happens if you try to defrag before the NVM is full.

EricB-ADI: It appears the baremetal implementation didn't have a bug fix. The code deletes sectors at a time, not bytes. The code just needed to be updated to not attempt to delete ```wsfNvm.availAddr``` worth of sectors

Libraries/Cordio/wsf/sources/targets/baremetal/wsf_nvm.c
```diff
@@ -402,14 +432,82 @@ bool_t WsfNvmEraseData(uint64_t id, WsfNvmCompEvent_t compCback)
...
+ *  \return  TRUE if defragment successful. FALSE otherwise
+ * 
+ *  \note   Defragment should only be called when the storage is full and a record has been invalidated.
+ *          Furthermore, copyBuf must be at least the size of WSF NVM allocated flash.
+ */
+/*************************************************************************************************/
+bool_t WsfNvmDefragment(uint8_t *copyBuf, uint32_t size)
+{
+#if WSF_ASSERT_ENABLED == 1
+    WSF_ASSERT(copyBuf && size >= wsfNvmCb.totalSize);
```
Jake-Carter, line 460: In general I would suggest checking for `WSF_ASSERT_ENABLED` _inside_ the `WSF_ASSERT` macro.  It cleans up the code and is much easier to maintain

Libraries/Cordio/wsf/sources/targets/freertos/wsf_nvm.c
Jake-Carter: @EricB-ADI ^^^

Jake-Carter (commented):
@sullivanmj if you can run a final test and give us confirmation your issue and findings have been resolved we'll merge it

@EricB-ADI in the meantime a very minor macro request below

Libraries/Cordio/wsf/sources/targets/baremetal/wsf_nvm.c
sullivanmj: @Jake-Carter Yes, with the changes provided by @EricB-ADI, the original issue can now be resolved on my end by detecting that the writing of pairing keys failed due to insufficient space, defragging, and then re-attempting to store the keys, which succeeds.

Thanks!

Jake-Carter (approved):

PR #990, opened 2024-04-15
chore(Other): Add zephyr related wrapper and system file in MSDK

ozersa:
### Description

hal_adi is repository that used by zephyr sdk. 
By this PR the content of this repository moved in MSDK repository to centralize it.

- Add wrappers files that used to provide common interface on zephyr shim driver.
- Add CMakefies that used to build drivers 
- Update zephyr-hal.sh script to copy new files to target repository (hal_adi).

Wrapper layer added to get one common zephyr shim driver on zephyr layer.
image

It will be nice if wrap layer function been added peripheral driver in future.

hal_adi folder layout listed below. After this PR being merged MAX folder content totally will comes from MSDK.
image

- [X] PR Title follows correct guidelines.

### DIFF
--- a/Libraries/zephyr/MAX/CMakeLists.txt
+++ b/Libraries/zephyr/MAX/CMakeLists.txt
@@ -1,8 +1,20 @@
-# Makefile - Analog Devices SDK
+##############################################################################
 #
-# Copyright (c) 2023 Analog Devices, Inc.
+# Copyright (C) 2023-2024 Analog Devices, Inc.
 #
-# SPDX-License-Identifier: Apache-2.0
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+##############################################################################
 
 
 if (${CONFIG_SOC} MATCHES "max32666")
--- a/Libraries/zephyr/MAX/CMakeLists.txt
+++ b/Libraries/zephyr/MAX/CMakeLists.txt
@@ -1,0 +1,50 @@
+# Makefile - Analog Devices SDK
+#
+# Copyright (c) 2023 Analog Devices, Inc.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+
+if (${CONFIG_SOC} MATCHES "max32666")
+    set(TARGET_LC "max32665")
+elseif (${CONFIG_SOC} MATCHES "max326651")
+    set(TARGET_LC "max32650")
+else ()
+    set(TARGET_LC ${CONFIG_SOC})
+endif()
+string(TOUPPER ${TARGET_LC} TARGET_UC)
+
+zephyr_compile_definitions(
+    -DTARGET=${TARGET_UC}
+    -DMSDK_NO_GPIO_CLK_INIT=1
+    -DMSDK_NO_LOCKING=1
+)
+
+get_filename_component(MSDK_LIBRARY_DIR "./Libraries" ABSOLUTE)
+set(MSDK_CMSIS_DIR       ${MSDK_LIBRARY_DIR}/CMSIS/Device/Maxim/${TARGET_UC})
+set(MSDK_PERIPH_DIR      ${MSDK_LIBRARY_DIR}/PeriphDrivers)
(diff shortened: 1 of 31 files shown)


Libraries/zephyr/MAX/Source/MAX32655/CMakeLists.txt
```diff
@@ -0,0 +1,171 @@
+# Makefile - Analog Devices SDK
```
lorne-maxim, line 1: Should the copyright/license text in these CMakeLists.txt files match the copyrights used in the other files?  Or do these files need the particular format you used?

lorne-maxim: @sihyung-maxim, what do you think?

sihyung-maxim: I believe this copyright/license text still works because the license identifier references Apache 2.0, and our `LICENSE` file at the root of the MSDK repo applies to every file and subdirectory within the root.

`.txt` files don't need to follow the same format as the source files as it depends on the context of how they're used - they can be used for documentation, data storing, configuration setting, and even as scripts/code. I think it's more of a matter of whether we want to keep consistency with our files, but I recommend using our default copyrights in this case since it's used in a scripting/make context.

ozersa: It was default copyright lines that we and other vendor added in zephyr files.
But it make sense it be consistent on the hal layer.
Updated please see: https://github.com/analogdevicesinc/msdk/pull/990/commits/632d57984a8d1eef6e426cfc1894203bde314881

sihyung-maxim (approved):

Libraries/zephyr/MAX/Source/MAX32655/CMakeLists.txt
MaureenHelm: Agree, be consistent with the rest of the MSDK

Libraries/zephyr/MAX/Include/wrap_max32xxx.h
Jake-Carter, line 1: @ozersa @MaureenHelm FYI we have `#include <mxc_device.h>` for each micro that we can use here to make maintenance of this file a bit easier.  It does the same job

(ex https://github.com/analogdevicesinc/msdk/blob/main/Libraries/PeriphDrivers/Include/MAX78000/mxc_device.h)

Libraries/zephyr/MAX/CMakeLists.txt
Jake-Carter, line 1: @ozersa does this add the source files to the top-level build?  Since Zephyr uses CMake it would also be nice to start setting things up more modularly for each library to support [find_package](https://cmake.org/cmake/help/latest/command/find_package.html)

Libraries/zephyr/MAX/CMakeLists.txt
```diff
@@ -0,0 +1,62 @@
...
+if (${CONFIG_SOC} MATCHES "max32666")
+    set(TARGET_LC "max32665")
+elseif (${CONFIG_SOC} MATCHES "max326651")
+    set(TARGET_LC "max32650")
+else ()
+    set(TARGET_LC ${CONFIG_SOC})
+endif()
+string(TOUPPER ${TARGET_LC} TARGET_UC)
+
+zephyr_compile_definitions(
```
Jake-Carter, line 29: @ozersa also suggest adding `-D${TARGET_UC}`

Libraries/zephyr/MAX/Include/wrap_max32xxx.h
ozersa: Thanks, due to we need to add extra addition as below in past we added a new files, now this file might be replaced with mxc_device.h. 
image

Libraries/zephyr/MAX/CMakeLists.txt
ozersa: Thanks, will consider on next update.

Libraries/zephyr/MAX/CMakeLists.txt
ozersa: Thanks noted.

Libraries/zephyr/MAX/CMakeLists.txt
ozersa: @Jake-Carter I took a look it now, to apply it on https://github.com/analogdevicesinc/msdk/pull/996 
But a little confused.
Do you propose replacing 
    -DTARGET=${TARGET_UC} with -DTARGET=**-D**${TARGET_UC} ?

Jake-Carter: @ozersa just to have both `TARGET` _and_ the actual value of `TARGET_UC` defined.

```CMake
zephyr_compile_definitions(
    -D${TARGET_UC}
    -DTARGET=${TARGET_UC}
    ...
)
```

Some places in the drivers need to check the target value.  Since the pre-processor can't do string comparisons sometimes `TARGET_UC` itself is checked.

```C
#ifdef MAX78000
// ...
#endif
```

PR #1004, opened 2024-04-29
fix(PeriphDrivers): Fix UART DMA Callbacks for UART1+

### LINKED ISSUE #998: In uart_reva.c, MXC_UART_RevA_Init does not initialize unused UARTs states
MXC_UART_RevA_Init only initializes the selected state's structure:
    // Initialize state struct
    unsigned int i = MXC_UART_GET_IDX((mxc_uart_regs_t *)uart);
    states[i].channelRx = -1;
    states[i].channelTx = -1;
    states[i].tx_req = NULL;
    states[i].rx_req = NULL;
    states[i].auto_dma_handlers = false;
Unused UART state's structures are not initiated.   The problem happens in MXC_UART_RevA_DMACallback when checking the channelTx:
    for (int i = 0; i < MXC_UART_INSTANCES; i++) {
        if (states[i].channelTx == ch) {
If it is not initialized, it thinks it is a valid state of a used UART and proceeds with checking the callback, which is also not initialized, and breaks from the loop, without iterating to the next UART instance. This results in not receiving the callbacks if UART1 is used (tested on MAX32665).

Jake-Carter:
### Description

Fixes #998 

As the internal UART state struct was only initialized per _used_ UART instance, attempting to use UART1 (or higher) with DMA auto handlers would never trigger the user's callback functions.  The state struct's DMA channel values would be initialized to 0 instead of -1, which the drivers would assume was a valid channel acquisition.

This PR fixes the bug by initializing the UART state array correctly.

### DIFF
--- a/Libraries/PeriphDrivers/Source/UART/uart_reva.c
+++ b/Libraries/PeriphDrivers/Source/UART/uart_reva.c
@@ -53,12 +53,10 @@ typedef struct {
 
 uart_reva_req_state_t states[MXC_UART_INSTANCES];
 
-#define DEFAULT_UART_REVA_REQ_STATE {\
-        .tx_req = NULL,\
-        .rx_req = NULL,\
-        .channelTx = -1,\
-        .channelRx = -1,\
-        .auto_dma_handlers = false\
+#define DEFAULT_UART_REVA_REQ_STATE                                       \
+    {                                                                     \
+        .tx_req = NULL, .rx_req = NULL, .channelTx = -1, .channelRx = -1, \
+        .auto_dma_handlers = false                                        \
     }
 
 bool g_is_state_initialized = false;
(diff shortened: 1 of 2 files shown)


EricB-ADI:
I recommend designated initializers instead of runtime. That way you don't need any runtime checks after and need to manage more global state.

```c
#define NUM_MEMBERS 4

typedef struct  
{
    int a , b, c;
}example_t;


example_t test[NUM_MEMBERS] = {
    [0 ... NUM_MEMBERS - 1] = {.a = -1, .b = -1, .c = -1}
};

```

Jake-Carter:
> I recommend designated initializers instead of runtime. That way you don't need any runtime operations after and need to manage more global state.
> 
> ```c
> #define NUM_MEMBERS 4
> 
> typedef struct  
> {
>     int32_t a , b, c;
> }example_t;
> 
> 
> example_t test[NUM_MEMBERS] = {
>     [0 ... NUM_MEMBERS - 1] = {.a = -1, .b = -1, .c = -1}
> };
> ```

Thanks @EricB-ADI, I was looking for this type of syntax.  Just updated the PR

Jake-Carter:
@EricB-ADI ready for final review

sihyung-maxim (approved):
Wow, never knew this syntax existed.

PR #1005, opened 2024-05-01
feat(PeriphDrivers): Add Basic MAX32657 PeriphDriver Support

Jake-Carter:
### Description

This PR will add basic PeriphDriver support for the ME30.  We are targeting a clean build.  Functional testing/fixes are beyond the scope of this PR.

I've added a MAX32657 Hello World as an initial build point.

**Status:**

- [x] Basic system implementation and top-level `mxc_sys.h`
- [ ] The SPI clock disable is defined in both PCLKDIS0 and PCLKDIS1.  Resolve in register files
- [ ] EXTCLK select is missing from gcr_regs.h.  Resolve in register files
- [ ] CLKCTRL sysclk_sel definitions seem incorrect.  Resolve in register files
- [x] Decide on secure vs non-secure default for top-level utility macros. (https://github.com/analogdevicesinc/msdk/pull/1005#issuecomment-2089264589)
- [x] Linkerfiles
- [x] Register-level builds
- [x] PeriphDriver Build
  - [x] mxc_assert
  - [x] mxc_delay
  - [x] mxc_lock
  - [x] mxc_sys
  - [ ] mxc_pins (should be updated when we get pin defs)
  - [x] AES
  - [x] CRC
  - [x] DMA
  - [x] FLC
  - [x] GPIO
  - [x] ICC
  - [ ] LP (missing register file)
  - [x] RTC
  - [x] SPI
  - [x] TMR
  - [x] UART
  - [x] WDT
  - [x] WUT


Jake-Carter:
@ozersa @ttmut @lorne-maxim @sihyung-maxim FYI.  See current status above

Jake-Carter:
@sihyung-maxim @lorne-maxim @ozersa @ttmut some updates below.

## 5/1/2024
I've updated the ME30 linker script (https://github.com/analogdevicesinc/msdk/pull/1005/commits/19787250196471f1ef0522253f960190fd08f5a8) and added a basic build mechanism for selecting between secure/non-secure builds (https://github.com/analogdevicesinc/msdk/pull/1005/commits/6f19ec6656398fc579716ed138ec1a50d50de07a).

The M33 will boot/reset into secure mode, so I've made it the default.  Additionally, access to the Flash controller is only possible in secure mode.

This allows the top-level max32657.h file to map the addresses to the correct aliases at build time (ex for UART: https://github.com/analogdevicesinc/msdk/pull/1005/commits/9f7bbfe13b9b115105a10fb7c70efe90a763bf26).

I think that this solves the issue of building _completely_ secure vs non-secure binaries.  Please comment on the approach before I apply it to the rest of the peripherals.

## Security Mode Transitions

Conceptually it is also possible to switch between the two modes at run-time.  However, if we link the entire application into the secure address mappings then I don't think it is possible to transition.  Even if we call a `BXNS`, all memory addresses will be linked into the secure space.  `BXNS` will only transition if the target memory address is in the non-secure space.

image
([Arm v8-M Architecture Reference Manual](https://developer.arm.com/documentation/ddi0553/latest/))

Conceptually, I have the following questions now: 
- **Is there a use-case/need for creating mixed security binaries?**
- **Is a driver-layer API for manually transitioning between the two modes ever useful?**

## DMA Complications

After updating the DMA drivers to account for the ME30's dual instances (https://github.com/analogdevicesinc/msdk/pull/1005/commits/df0e56640da75070e9561fed270a9d16dbe6ee7f) there are some complications.

Many DMA drivers do not expose the DMA instance as an argument at the top-level header.  The bottom-level RevA/RevB files call these variations directly.  Ex attempting to build AES:

```C
- CC /home/jhcarter/repos/msdk/Libraries/CMSIS/../PeriphDrivers/Source/AES/aes_revb.c
/home/jhcarter/repos/msdk/Libraries/CMSIS/../PeriphDrivers/Source/AES/aes_revb.c: In function 'MXC_AES_RevB_TXDMAConfig':
/home/jhcarter/repos/msdk/Libraries/CMSIS/../PeriphDrivers/Source/AES/aes_revb.c:243:5: error: too few arguments to function 'MXC_DMA_Init'
  243 |     MXC_DMA_Init();
```

On the ME30, DMA1 is a _secure_ DMA with access to the secure space, and DMA0 is a _non-secure_ instance.  If a completely secure image is built, DMA0 will be useless.  So - **should we treat the ME30 as if it has a single DMA instance, and assign DMA1/DMA0 based on secure vs non-secure builds?**

sihyung-maxim:
> @sihyung-maxim @lorne-maxim @ozersa @ttmut some updates below.
> 
> ## 5/1/2024
> I've updated the ME30 linker script ([1978725](https://github.com/analogdevicesinc/msdk/commit/19787250196471f1ef0522253f960190fd08f5a8)) and added a basic build mechanism for selecting between secure/non-secure builds ([6f19ec6](https://github.com/analogdevicesinc/msdk/commit/6f19ec6656398fc579716ed138ec1a50d50de07a)).
> 
> The M33 will boot/reset into secure mode, so I've made it the default. Additionally, access to the Flash controller is only possible in secure mode.
> 
> This allows the top-level max32657.h file to map the addresses to the correct aliases at build time (ex for UART: [9f7bbfe](https://github.com/analogdevicesinc/msdk/commit/9f7bbfe13b9b115105a10fb7c70efe90a763bf26)).
> 
> I think that this solves the issue of building _completely_ secure vs non-secure binaries. Please comment on the approach before I apply it to the rest of the peripherals.
> 
> ## Security Mode Transitions
> Conceptually it is also possible to switch between the two modes at run-time. However, if we link the entire application into the secure address mappings then I don't think it is possible to transition. Even if we call a `BXNS`, the rest of the code will be linked into the secure space. So we've effectively locked out the application from that point on.
> 
> image ([Arm v8-M Architecture Reference Manual](https://developer.arm.com/documentation/ddi0553/latest/))
> 
> Conceptually, I have the following questions now:
> 
> * **Is there a use-case/need for creating mixed security binaries?**
> * **Is a driver-layer API for manually transitioning between the two modes ever useful?**
> 
> ## DMA Complications
> After updating the DMA drivers to account for the ME30's dual instances ([df0e566](https://github.com/analogdevicesinc/msdk/commit/df0e56640da75070e9561fed270a9d16dbe6ee7f)) there are some complications.
> 
> Many DMA drivers do not expose the DMA instance as an argument at the top-level header. The bottom-level RevA/RevB files call these variations directly. Ex attempting to build AES:
> 
> ```c
> - CC /home/jhcarter/repos/msdk/Libraries/CMSIS/../PeriphDrivers/Source/AES/aes_revb.c
> /home/jhcarter/repos/msdk/Libraries/CMSIS/../PeriphDrivers/Source/AES/aes_revb.c: In function 'MXC_AES_RevB_TXDMAConfig':
> /home/jhcarter/repos/msdk/Libraries/CMSIS/../PeriphDrivers/Source/AES/aes_revb.c:243:5: error: too few arguments to function 'MXC_DMA_Init'
>   243 |     MXC_DMA_Init();
> ```
> 
> On the ME30, DMA1 is a _secure_ DMA with access to the secure space, and DMA0 is a _non-secure_ instance. If a completely secure image is built, DMA0 will be useless. So - **should we treat the ME30 as if it has a single DMA instance, and assign DMA1/DMA0 based on secure vs non-secure builds?**

I'm still working on my side of the updates, but to answer your DMA question:

Yeah, I think we should. We still need to define separate names for the DMAs in the registers: for example, GCR_PCLKDISn and _RSTn registers has fields representing each DMA instance. Maybe a better name for the two DMA instances are: `DMA0 -> DMA_NS` and `DMA1 -> to DMA_S`. Then as you said, define what the undecorated DMA name would be depending on the secure vs non-secure builds. So it would look something like this:

In max32657.h:
```
... Interrupt number table
    DMA_NS_CH0_IRQn,
    DMA_NS_CH1_IRQn,
    DMA_NS_CH2_IRQn,
    DMA_NS_CH3_IRQn,
    DMA_S_CH0_IRQn,
    DMA_S_CH1_IRQn,
    DMA_S_CH2_IRQn,
    DMA_S_CH3_IRQn,
...
... DMA section
    #define MXC_BASE_DMA_NS ((uint32_t)0x40028000UL)
    #define MXC_DMA_NS ((mxc_dma_regs_t *)MXC_BASE_DMA_NS)

    #define MXC_BASE_DMA_S ((uint32_t)0x50028000UL)
    #define MXC_DMA_S ((mxc_dma_regs_t *)MXC_BASE_DMA_S)

#if defined(MSECURITY_MODE_SECURE)
    #define MXC_BASE_DMA MXC_BASE_DMA_S
    #define MXC_DMA MXC_DMA_S
    #define DMA_CH0_IRQn DMA_S_CH0_IRQn
    #define DMA_CH1_IRQn DMA_S_CH1_IRQn
    #define DMA_CH2_IRQn DMA_S_CH2_IRQn
    #define DMA_CH3_IRQn DMA_S_CH3_IRQn
    #define DMA_CH0_IRQHandler DMA_S_CH0_IRQHandler
    #define DMA_CH1_IRQHandler DMA_S_CH1_IRQHandler
    #define DMA_CH2_IRQHandler DMA_S_CH2_IRQHandler
    #define DMA_CH3_IRQHandler DMA_S_CH3_IRQHandler
#else
    #define MXC_BASE_DMA MXC_BASE_DMA_NS
    #define MXC_DMA MXC_DMA_NS
    #define DMA_CH0_IRQn DMA_NS_CH0_IRQn
    #define DMA_CH1_IRQn DMA_NS_CH1_IRQn
    #define DMA_CH2_IRQn DMA_NS_CH2_IRQn
    #define DMA_CH3_IRQn DMA_NS_CH3_IRQn
    #define DMA_CH0_IRQHandler DMA_NS_CH0_IRQHandler
    #define DMA_CH1_IRQHandler DMA_NS_CH1_IRQHandler
    #define DMA_CH2_IRQHandler DMA_NS_CH2_IRQHandler
    #define DMA_CH3_IRQHandler DMA_NS_CH3_IRQHandler
#endif
```

Also, I've been meaning to implement this name change for newer parts, but we can rename the DMA vector names to something more accurate.
```
   DMA0_IRqn -> DMA_CH0_IRQn
   DMA1_IRqn -> DMA_CH1_IRQn
   DMA2_IRqn -> DMA_CH2_IRQn
   DMA3_IRqn -> DMA_CH3_IRQn
```
The channel number wouldn't be where you would expect the DMA instance number to be. What do we think?

Jake-Carter:
@sihyung-maxim I think this solution of mapping onto secure/non-secure base address based on the build value makes sense and would work well. 

I'm reading more about TrustZone and the SAU/MPCs/PPCs and it sounds like we can partition the memory spaces and security attributes for each peripheral individually at startup.

The standard way to do things is to build secure applications separately (first), then compile and link in the non-secure image into the same binary afterwards.  So I'm working on a mechanism now for that similar to what we've done with the dual-core micros.  The _secure_ image will be responsible for initializing the security attributes for each peripheral on startup.  Then the non-secure image will be appended onto the secure image's SRAM/Flash using the _non-secure_ address aliases.

This will turn the ME30 projects into dual-project workspaces for the most-part, but seems to be the standard approach.

So this works well.

> Also, I've been meaning to implement this name change for newer parts, but we can rename the DMA vector names to something more accurate.

> ```C
>   DMA0_IRqn -> DMA_CH0_IRQn
>   DMA1_IRqn -> DMA_CH1_IRQn
>   DMA2_IRqn -> DMA_CH2_IRQn
>   DMA3_IRqn -> DMA_CH3_IRQn
>```
> The channel number wouldn't be where you would expect the DMA instance number to be. What do we think?

Having the channel is good.  I think having the instance in there is also needed.  I went with this for the header file atm:

```C
DMA0_CH0_IRQn,          /* 0x20  0x0080  32: DMA0 Channel 0 */
DMA0_CH1_IRQn,          /* 0x21  0x0084  33: DMA0 Channel 1 */
DMA0_CH2_IRQn,          /* 0x22  0x0088  34: DMA0 Channel 2 */
DMA0_CH3_IRQn,          /* 0x23  0x008C  35: DMA0 Channel 3 */
DMA1_CH0_IRQn,          /* 0x24  0x0090  36: DMA1 Channel 0 */
DMA1_CH1_IRQn,          /* 0x25  0x0094  37: DMA1 Channel 1 */
DMA1_CH2_IRQn,          /* 0x26  0x0098  38: DMA1 CHannel 2 */
DMA1_CH3_IRQn,          /* 0x27  0x009C  39: DMA1 Channel 3 */
```

Jake-Carter:
@sihyung-maxim there is also the concept of secure vs non-secure interrupts.  We will need to add some additional NVIC entries with a secure/non-secure offset

sihyung-maxim (approved):

Jake-Carter:
Merged.  

The build system can specify secure (default) vs non-secure builds with the new `MSECURITY_MODE` [build config variable](https://analogdevicesinc.github.io/msdk/USERGUIDE/#build-configuration-variables).  Acceptable values are:
- `SECURE`
- `NONSECURE`

I recommend setting this in `project.mk` so that any changes to its value trigger automatic full rebuilds.

**Note we still need to check/update pin definitions**

PR #1008, opened 2024-05-03
feat(CMSIS): Update MAX32657 CMSIS with TrustZone support

sihyung-maxim:
### Description

Some updates to the CMSIS for Verification.
- Added undecorated MXC_BASE_* definitions
- Removed `GET_S_` and `GET_NS` definitions from max32657.h
- Define both DMA instances
- Added partition_max32657.h (from CMSIS) to handle SAU memory region partitioning for secure and non-secure code.
- Updated linker, startup, and system files (SystemInit, SAU, Stack Sealing, Interrupt Tables)

Both DMA0 and DMA1 need to be defined, so we can't use the single `MXC_DMA` instance anymore.
1. DMA0 has a secure and non-secure mapping - accessible for secure and non-secure builds. However, DMA0 can not access secure memory or secure peripherals.
2. DMA1 only has a secure mapping - accessible for secure builds. But DMA1 can access both secure or non-secure memory and peripherals.

I'll have a separate PR for the remaining SYS register updates that I need to make.

Edit: Removing multiple linkers comment, just saw your changes @Jake-Carter. But I don't think we need multiple vector table definitions and startup_max32657.S startup code.

We can have one vector table defined that includes all secure and non-secure handlers, and we can dynamically switch between secure and non-secure vector tables during startup process using the `SCB->VTOR` vector table offset register. The `SCB->VTOR` register is banked into two parts, and the appropriate offset for secure and non-secure is used depending on the security context. We can also explicitly label the secure-only interrupts with the `_S` names, and include a build warning/error if the non-secure application references a secure handler.

The secure vector table is set up in `SystemInit()` called from Reset_Handler on startup. We'd have to figure out a mechanism for non-secure builds (non-secure reset handler) to set up the non-secure vector table (using `SCB->VTOR` again). This is at least my current understanding of how we might go about this for our bare metal MSDK.

There's a lot to think about - but given our schedule and priorities, we can focus on handling the secure mode for now so verification can begin. These changes are geared towards finishing the Secure build system changes that you've made, and to prep the non-secure builds for down the road.

### DIFF
--- a/Libraries/CMSIS/Device/Maxim/MAX32657/Include/max32657.h
+++ b/Libraries/CMSIS/Device/Maxim/MAX32657/Include/max32657.h
@@ -625,7 +625,11 @@ We may want to handle GET_IRQ better...
 /******************************************************************************/
 /*                                                                        DMA */
 #define MXC_DMA_CHANNELS (4)
+#if IS_SECURE_ENVIRONMENT
 #define MXC_DMA_INSTANCES (2)
+#else
+#define MXC_DMA_INSTANCES (1)
+#endif
 
 /* Non-secure Mapping */
 #define MXC_BASE_DMA0_NS ((uint32_t)0x40028000UL)
--- a/Libraries/CMSIS/Device/Maxim/MAX32657/Include/max32657.h
+++ b/Libraries/CMSIS/Device/Maxim/MAX32657/Include/max32657.h
@@ -59,7 +59,6 @@
 /* ================================================================================ */
 
 // clang-format off
-// TODO(ME30): Secure vs non-secure interrupt vectors
 typedef enum {
     NonMaskableInt_IRQn = -14,
     HardFault_IRQn = -13,
@@ -97,10 +96,10 @@ typedef enum {
     DMA0_CH1_IRQn,          /* 0x21  0x0084  33: DMA0 Channel 1 */
     DMA0_CH2_IRQn,          /* 0x22  0x0088  34: DMA0 Channel 2 */
     DMA0_CH3_IRQn,          /* 0x23  0x008C  35: DMA0 Channel 3 */
-    DMA1_CH0_IRQn,          /* 0x24  0x0090  36: DMA1 Channel 0 */
-    DMA1_CH1_IRQn,          /* 0x25  0x0094  37: DMA1 Channel 1 */
-    DMA1_CH2_IRQn,          /* 0x26  0x0098  38: DMA1 CHannel 2 */
-    DMA1_CH3_IRQn,          /* 0x27  0x009C  39: DMA1 Channel 3 */
+    DMA1_CH0_IRQn,          /* 0x24  0x0090  36: DMA1 Channel 0 (Secure) */
+    DMA1_CH1_IRQn,          /* 0x25  0x0094  37: DMA1 Channel 1 (Secure) */
+    DMA1_CH2_IRQn,          /* 0x26  0x0098  38: DMA1 CHannel 2 (Secure) */
+    DMA1_CH3_IRQn,          /* 0x27  0x009C  39: DMA1 Channel 3 (Secure) */
     WUT0_IRQn,              /* 0x28  0x00A0  40: Wakeup Timer 0 */
     WUT1_IRQn,              /* 0x29  0x00A4  41: Wakeup TImer 1 */
     GPIOWAKE_IRQn,          /* 0x2A  0x00A8  42: GPIO Wakeup */
@@ -233,8 +232,10 @@ typedef enum {
 #define MXC_SIR_S ((mxc_sir_regs_t *)MXC_BASE_SIR_S)
 
 #if IS_SECURE_ENVIRONMENT
+#define MXC_BASE_SIR MXC_BASE_SIR_S
 #define MXC_SIR MXC_SIR_S
 #else
+#define MXC_BASE_SIR MXC_BASE_SIR_NS
 #define MXC_SIR MXC_SIR_NS
 #endif
 
@@ -250,8 +251,10 @@ typedef enum {
 #define MXC_FCR_S ((mxc_fcr_regs_t *)MXC_BASE_FCR_S)
 
 #if IS_SECURE_ENVIRONMENT
+#define MXC_BASE_FCR MXC_BASE_FCR_S
 #define MXC_FCR MXC_FCR_S
 #else
+#define MXC_BASE_FCR MXC_BASE_FCR_NS
 #define MXC_FCR MXC_FCR_NS
 #endif
 
@@ -268,8 +271,10 @@ typedef enum {
 #define MXC_WDT_S ((mxc_wdt_regs_t *)MXC_BASE_WDT_S)
 
 #if IS_SECURE_ENVIRONMENT
+#define MXC_BASE_WDT MXC_BASE_WDT_S
 #define MXC_WDT MXC_WDT_S
 #else
+#define MXC_BASE_WDT MXC_BASE_WDT_NS
 #define MXC_WDT MXC_WDT_NS
 #endif
 
@@ -278,15 +283,17 @@ typedef enum {
 
 /* Non-secure Mapping */
 #define MXC_BASE_SVM_NS ((uint32_t)0x40004800UL)
-#define MXC_SVM_NS //TODO(ME30): Add SVM controller registers.
+#define MXC_SVM_NS 0 //TODO(ME30): Add SVM controller registers.
 
 /* Secure Mapping */
 #define MXC_BASE_SVM_S ((uint32_t)0x50004800UL)
-#define MXC_SVM_S //TODO(ME30): Add SVM controller registers.
+#define MXC_SVM_S 0 //TODO(ME30): Add SVM controller registers.
 
 #if IS_SECURE_ENVIRONMENT
+#define MXC_BASE_SVM MXC_BASE_SVM_S
 #define MXC_SVM MXC_SVM_S //TODO(ME30): Add SVM controller registers
 #else
+#define MXC_BASE_SVM MXC_BASE_SVM_NS
 #define MXC_SVM MXC_SVM_NS
 #endif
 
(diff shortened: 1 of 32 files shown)


Libraries/CMSIS/Device/Maxim/MAX32657/Include/max32657.h
```diff
@@ -601,52 +615,70 @@ We may want to handle GET_IRQ better...
...
 #define MXC_I3C MXC_I3C_NS
 #endif
 
 /******************************************************************************/
 /*                                                                        DMA */
 #define MXC_DMA_CHANNELS (4)
-#define MXC_DMA_INSTANCES (1)
-// ^ Note: We have 2 DMA instances in hardware, but they are secure vs non-secure
-// instances.  Therefore we treat the part as if there is only 1.
+#define MXC_DMA_INSTANCES (2)
```
Jake-Carter, line 628: If you try to build with this you'll see an issue at the bottom-level DMA driver file with the handler definitions.  Also RevA depends on `MXC_DMA`

Libraries/CMSIS/Device/Maxim/MAX32657/Include/max32657.h
```diff
@@ -601,52 +615,70 @@ We may want to handle GET_IRQ better...
...
-/* DMA0 instance only for secure mode. */
 
 /* Secure Mapping */
 // TODO(ME30): Is there actuall a secure mapping for DMA0?
+//             -Yes, DMA0 can be accessed from secure mode. Realizing this, I think
+//                  we would still have to define two DMA instances.
+//                  DMA0 can only access the non-secure mappings of the peripherals,
+//                  but DMA0 can be accessed in both Non-secure and Secure code.
+//                  DMA1 can access both secure and non-secure addresses of the peripherals,
+//                  but DMA1 can Only be accessed in Secure code.
```
Jake-Carter, line 641: DMA0 can be _accessed_ from secure mode but my understanding is that the controller cannot access secure memory/peripherals.  I think we should still preserve the ability to access it at the register-level, but for this first pass at the drivers I think we should revert back to the previous setup

Libraries/CMSIS/Device/Maxim/MAX32657/Include/max32657.h
```diff
@@ -601,52 +615,70 @@ We may want to handle GET_IRQ better...
...
                  ((i) == 2) ? DMA1_CH2_IRQn : \
                  ((i) == 3) ? DMA1_CH3_IRQn : \
                               0))
+
 #else
-#define MXC_DMA MXC_DMA0_NS
-#define MXC_DMA_CH_GET_IRQ(i)             \
+#define MXC_BASE_DMA0 MXC_BASE_DMA0_NS
+#define MXC_DMA0 MXC_DMA0_NS
+// TODO(DMA1): Not entirely show how to handle access to MXC_DMA1 in non-secure mode.
```
Jake-Carter, line 663: The previous setup solves this problem as well

Libraries/CMSIS/Device/Maxim/MAX32657/Source/GCC/max32657.ld
```diff
@@ -61,19 +61,20 @@ SECTIONS {
...
+     * Secure Gatway (SG) veneers.
+     *  All SG veneers are placed in the special output section .gnu.sgstubs.
+     */
+    .gnu.sgstubs :
     {
-        *(.ARM.extab* .gnu.linkonce.armextab.*)
+        _sg_veneers = .;
+        KEEP(*(.gnu.sgstubs*))
+        _esg_veneers = .;
     } > FLASH_S
```
Jake-Carter, line 73: See https://github.com/analogdevicesinc/msdk/commit/4e9156e999a23ed16cbaa7125005362013771573

I updated the linkerfile to support secure/non-secure from the same file.  I think we should be able to wrap this section in the same way so that it only gets generated for the secure builds.

Libraries/CMSIS/Device/Maxim/MAX32657/Source/GCC/startup_max32657.S
```diff
@@ -91,6 +154,19 @@ Reset_Handler:
     ldr r0, =__StackTop
     mov sp, r0
 
+#if IS_SECURE_ENVIRONMENT
```
Jake-Carter, line 157: `IS_SECURE_ENVIRONMENT` is defined in max32657.h - not sure if this will work in the assembly file

Libraries/CMSIS/Device/Maxim/MAX32657/Source/system_max32657.c
```diff
@@ -56,15 +59,12 @@ __weak void SystemCoreClockUpdate(void)
...
     case MXC_S_GCR_CLKCTRL_SYSCLK_SEL_IBRO:
         base_freq = IBRO_FREQ;
         break;
-    case MXC_S_GCR_CLKCTRL_SYSCLK_SEL_ERTCO:
-        base_freq = ERTCO_FREQ;
-        break;
-    // case MXC_S_GCR_CLKCTRL_SYSCLK_SEL_EXTCLK:
-    //     base_freq = EXTCLK_FREQ;
+    // case MXC_S_GCR_CLKCTRL_SYSCLK_SEL_ERTCO:
+    //     base_freq = ERTCO_FREQ;
```
Jake-Carter, line 67: Removed EXTCLK and commented out ERTCO?

Libraries/CMSIS/Device/Maxim/MAX32657/Source/system_max32657.c
```diff
@@ -121,31 +122,42 @@ __weak int Board_Init(void)
...
 
-    /* Make sure interrupts are enabled. */
-    __enable_irq();
+#if (__FPU_PRESENT == 1U)
+    /* Enable FPU - coprocessor slots 10 & 11 full access */
+    SCB->CPACR |= SCB_CPACR_CP10_Msk | SCB_CPACR_CP11_Msk;
+#endif /* __FPU_PRESENT check */
 
-    /* Enable instruction cache */
-    MXC_ICC_Enable(MXC_ICC);
```
Jake-Carter, line 135: ICC enable removed?

Libraries/CMSIS/Device/Maxim/MAX32657/Include/partition_max32657.h
Jake-Carter, line 1: Do you know how we created this file/where it came from?

Jake-Carter (changes requested):

Libraries/CMSIS/Device/Maxim/MAX32657/Include/max32657.h
sihyung-maxim: Yes, that is why I added a WIP for this PR before committing to this.

Libraries/CMSIS/Device/Maxim/MAX32657/Include/max32657.h
sihyung-maxim: Yes, you are correct - as mentioned in the PR description above:

> Both DMA0 and DMA1 need to be defined, so we can't use the single `MXC_DMA` instance anymore.
> 
> 1. DMA0 has a secure and non-secure mapping - accessible for secure and non-secure builds. However, DMA0 can not access secure memory or secure peripherals.
> 2. DMA1 only has a secure mapping - accessible for secure builds. But DMA1 can access both secure or non-secure memory and peripherals.

The diction for all the security can get confusing. If we do a secure build, there needs to be an option to access both DMA0 and DMA1.

Libraries/CMSIS/Device/Maxim/MAX32657/Source/system_max32657.c
sihyung-maxim: My bad, oversight on my part.

Libraries/CMSIS/Device/Maxim/MAX32657/Include/partition_max32657.h
sihyung-maxim: Arm provides it through their CMSIS-Core template from GitHub: https://github.com/ARM-software/CMSIS_5.

Some vendors don't use this partition file and rather set the SAU registers themselves somewhere for finer grain control. However, at the moment, it was decided to use the template to simplify handling for this first pass.

Libraries/CMSIS/Device/Maxim/MAX32657/Source/system_max32657.c
sihyung-maxim: Yep, I didn't want to add it for the initial pass for the verification and roadtest teams. It is a new internal cache IP (well a superset of our existing one). We can add it for increased speedups, but we should do some testing for this new IP before we enable it by default as there have been some cache issues in the past.

Libraries/CMSIS/Device/Maxim/MAX32657/Include/max32657.h
```diff
@@ -233,8 +232,10 @@ typedef enum {
 #define MXC_SIR_S ((mxc_sir_regs_t *)MXC_BASE_SIR_S)
 
 #if IS_SECURE_ENVIRONMENT
+#define MXC_BASE_SIR MXC_BASE_SIR_S
```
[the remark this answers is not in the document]
sihyung-maxim: Not in the SDK but I added them for the verification and roadtest teams. We're keeping the undecorated names that were previously accessible. The codebase for the roadtest and verification environments is dependent on the existing convention from our CMSIS files.

Libraries/CMSIS/Device/Maxim/MAX32657/Source/GCC/startup_max32657.S
sihyung-maxim: Yeah, there are unfinished parts of this PR that I was unsure about. I figure we would need an assembly project flag defined for this when a secure build happens.

Jake-Carter: `PROJ_AFLAGS += IS_SECURE_ENVIRONMENT` added to the secure section of gcc.mk would work well

Libraries/CMSIS/Device/Maxim/MAX32657/Source/system_max32657.c
Jake-Carter: Ok, maybe comment it out and add a TODO so we don't forget?

sihyung-maxim: Good call.

Libraries/PeriphDrivers/Source/DMA/dma_reva.c
```diff
@@ -500,7 +500,11 @@ int MXC_DMA_RevA_MemCpy(mxc_dma_reva_regs_t *dma, void *dest, void *src, int len
         return retval;
     }
 
+#if (TARGET_NUM == 32657)
```
Jake-Carter, line 503: :(

sihyung-maxim: Yep, agreed. I don't like this either lol. We can change this in another separate PR.

Jake-Carter (approved):

PR #1022, opened 2024-05-22
feat(BLE): Added User Configurable Duplicate Filtering

BrentK-ADI:
Updated the Extended scan functionality to allow the user to specify if duplicate packets should be filtered during the scanning process.  This was previously hard coded to TRUE.  The legacy Scan function remains hard coded for simplicity of use and compatibility with existing code bases.

All existing BLE example applications use the legacy AppScanStart call. This was tested against a customer project using the AppExtScanStart capability.


EricB-ADI (approved):

Jake-Carter:
@EricB-ADI good to merge?

PR #1023, opened 2024-05-23
feat(CMSIS,Examples,PeriphDrivers)!: Add MAX32665 Dual Core example and Core 1 CMSIS updates

sihyung-maxim:
### Description

This change only impacts the MAX32665.

This PR organizes the CMSIS files for Core 1 and adds a dual core example. Core1 was not properly set up when the MAX32665 SDK files were initially committed into the MSDK.

There may be some breaking changes because some of the CORE1 files were relocated - I deprecated the old Core1 files that were moved to CMSIS. I also updated some of the function names to be more consistent - any duplicate core0 function name that needs to be supported for core1 will use the `_Core1` suffix.


EricB-ADI:
MAX32665 isn't covered by the ble examples workflow. Just ignore the issue.

sihyung-maxim:
> just for a note: this PR only adds the example for the MAX32666 EVKIT [514fc1a](https://github.com/analogdevicesinc/msdk/commit/514fc1a954de0a4138dbeffa1b019bc429b3ec74)
> 
> Shouldn't it be added the the FTHR board as well?

Have not tested this on the featherboard yet (but it should still work). I'll test on the featherboard and then update its board files for it.

Examples/MAX32665/Dual_Core/project.mk
```diff
@@ -0,0 +1,20 @@
...
+
+# **********************************************************
+
+# Add your config here!
+
+# Build with the necessary Core1 startup/system files.
+ARM_DUALCORE=1
+
+# Separate directories for Core 0 and Core 1 code.
+VPATH += Core0
```
Jake-Carter, line 16: Might not be necessary to separate the directories since both cores share the same arch.  You could even define `main` and `main_core1` in the same file.

Jake-Carter (approved):

Examples/MAX32665/Dual_Core/project.mk
sihyung-maxim: Yeah, mainly did this so users can organize their code between the two cores.

sihyung-maxim:
Works on FTHR.

PR #1025, opened 2024-05-27
fix(PeriphDrivers): Add CSI Capture Timeout

isztldav:
It can happen that the g_frame_complete flag is never set, thus the device gets stuck in an endless loop. This is now checked by a timing constraint


Jake-Carter (commented):
Thanks @isztldav, this is a good addition.

The implementation looks clean, and I'll approve this after a quick functional test on my side

Jake-Carter (approved):

PR #1041, opened 2024-06-11
fix(PeriphDrivers): Fix SDHC Peripheral Failure for Alternative System Clocks

OzgunKanal:
### Description

According to the UG, SDHC peripheral is driven by always 96 MHz. Therefore, using System clock in the function MXC_SDHC_Get_Input_Clock_Freq results failure in SDHC peripheral when SytemClock is configured other than 96 MHz. Moreover, according to the UG the SDHC is driven by the 96 MHz Internal High-Speed Oscillator. Similar situations occur for ME10, ME13 and AI87 and commits relating to these MCUs will be pushed after this commit, discussions and related tests.

### DIFF
--- a/Libraries/CMSIS/Device/Maxim/MAX32650/Include/system_max32650.h
+++ b/Libraries/CMSIS/Device/Maxim/MAX32650/Include/system_max32650.h
@@ -46,8 +46,8 @@ extern "C" {
 #define NANORING_FREQ 8000
 #endif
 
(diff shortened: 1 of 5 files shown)


Jake-Carter:
Thanks @OzgunKanal, agreed we can use the oscillator frequency macros for this instead of SystemCoreClock.  I think this change is good to apply to the other micros as well.

Jake-Carter (approved):

PR #1045, opened 2024-06-11
feat(MiscDrivers): Add API for Setting SDHC Data Width

EricB-ADI:
### Description

Added API to set default SPI mode (single/quad) for SDHC. Compile time default set to single mode to maintain backwards compatibility.


Libraries/CMSIS/Device/Maxim/GCC/mxc_version.mk
Jake-Carter, line 1: Don't delete this

mxc_version.h
Jake-Carter, line 1: or this

Jake-Carter (commented):
Some feedback below - take it or leave it.  This looks good, just some thoughts on a potentially cleaner diff for this.

On the version number files - See #937.  I still need to get around to an automatic workflow for updating them, but don't delete them in the meantime

Libraries/SDHC/Source/sdhc_lib.c
```diff
@@ -236,6 +237,8 @@ int MXC_SDHC_Lib_SetBusWidth(mxc_sdhc_data_width bus_width)
     uint32_t card_status;
     int result;
 
+    bus_width = bus_width == MXC_SDHC_LIB_DEFAULT_DATA ? MXC_SDHC_Get_Default_DataWidth() : bus_width;
```
Jake-Carter, line 240: I think it might be cleaner to come in "over the top" and avoid adding the extra enum value.  A direct getter/setter to the state variable would work, so your call from `diskio.c` would look like:

```C
MXC_SDHC_Lib_Write(sector, (void *)buff, count, MXC_SDHC_Get_DataWidth())
```

Usually you're not gonna change your data width more than once at run-time, so a mechanism for setting the default value might be better implemented as a build-time option.  Users can use the setter API you've exposed on initialization.

This would let us avoid these modifications to `sdhc_lib.c`, since the data width is already an arg to these functions

Jake-Carter (approved):

PR #1016, opened 2024-05-13
fix(PeriphDrivers): Fix `MXC_SYS_GetUSN` Buffer Overflow

Jake-Carter:
### Description

Fixes #1006 

This PR updates `MXC_SYS_GetUSN` to prevent buffer overflow.  The application-side buffer should be `MXC_SYS_USN_LEN`, but we were using `MXC_SYS_USN_CHECKSUM_LEN` which is a larger buffer size with some additional padding used for the internal CRC calculation.

### Checklist Before Requesting Review

- [ ] PR Title follows correct guidelines.
- [ ] Description of changes and all other relevant information.
- [ ] (Optional) Link any related GitHub issues [using a keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)
- [ ] (Optional) Provide info on any relevant functional testing/validation.  For API changes or significant features, this is not optional.

PR #1050, opened 2024-06-20
fix(PeriphDrivers): MXC_SYS_GetUSN Checksum error

EricB-ADI:
### Description
For some MCUs, the ``MXC_SYS_GetUSN`` function used the AES peripheral to calculate the checksum. 
#1016 Addresses an issue with the possibility of an unbounded memory access. The ``MXC_SYS_GetUSN`` function now only copies and clears the 13 bytes of the USN. However, the checksum is calculated using the USN checksum length not the USN length, so the CRC was calculated using the correct 13 bytes of the USN plus 3 uninitialized bytes causing checksum errors. This patch fixes this issue.

### DIFF
--- a/Libraries/PeriphDrivers/Source/SYS/sys_me18.c
+++ b/Libraries/PeriphDrivers/Source/SYS/sys_me18.c
@@ -97,9 +97,9 @@ int MXC_SYS_GetUSN(uint8_t *usn, uint8_t *checksum)
         /* Initialize key and plaintext */
         memset(key, 0, MXC_SYS_USN_CHECKSUM_LEN);
         memset(pt32, 0, MXC_SYS_USN_CHECKSUM_LEN);
-        memcpy(pt32, usn, MXC_SYS_USN_CHECKSUM_LEN);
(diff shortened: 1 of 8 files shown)


Libraries/PeriphDrivers/Source/SYS/sys_me18.c
```diff
@@ -97,9 +97,9 @@ int MXC_SYS_GetUSN(uint8_t *usn, uint8_t *checksum)
         /* Initialize key and plaintext */
         memset(key, 0, MXC_SYS_USN_CHECKSUM_LEN);
         memset(pt32, 0, MXC_SYS_USN_CHECKSUM_LEN);
-        memcpy(pt32, usn, MXC_SYS_USN_CHECKSUM_LEN);
+        memcpy(pt32, usn, MXC_SYS_USN_LEN);
 
-        /* Read the checksum from the info block */
+        /* Read the checksum from the inspfo block */
```
sihyung-maxim, line 102: typo?

sihyung-maxim (approved):

PR #1052, opened 2024-06-21
perf(Other): Use FreeRTOS Heap Allocation Scheme 4, Allow Changing FreeRTOS Heap Allocation Scheme 

Brandon-Hurst:
### Description

FreeRTOS gives some flexibility in the memory allocation scheme used to allow optimizing size/speed/heap fragmentation risk. The FreeRTOS documentation indicates that _heap_2.c is considered outdated_, and that _applications should at a minimum use heap_4 as an alternative_. Heap 4 uses a smarter allocation algorithm for avoiding heap fragmentation (first-fit algorithm + aggregates free RAM segments). 

Sources for info: 
https://www.freertos.org/a00111.html
https://github.com/FreeRTOS/FreeRTOS-Kernel-Book/blob/main/ch03.md

#### Summary of Changes
- Changed default FreeRTOS heap allocation to heap_4.c -heap_4 is recommended by FreeRTOS as an updated alternate to heap_2 for avoiding heap fragmentation 
- Added **FREERTOS_HEAP_TYPE** compile time macro to select the heap allocation method 
  - Default value is 4, but can be 1-5. Error message is included if invalid input is given. 
- Added an entry for FREERTOS_HEAP_TYPE in MAX78000's FreeRTOSDemo project.mk

This would (positively) affect the way FreeRTOS builds for all supported MCUs. 

#### Tests
Tested library builds with all 5 different heap types using FreeRTOSDemo for MAX78000 as the test source code. Example builds with the appropriate heap files based on the new FREERTOS_HEAP_TYPE macro. If the macro is not defined in project.mk, the library builds with heap_4.c by default. 

### DIFF
--- a/Examples/MAX78000/FreeRTOSDemo/project.mk
+++ b/Examples/MAX78000/FreeRTOSDemo/project.mk
@@ -16,6 +16,11 @@ DEBUG=1
 
 LIB_FREERTOS = 1
 
+# Can provide a value for the FREERTOS heap allocation scheme
+# Default value is 4
+# FREERTOS_HEAP_TYPE := 2
+# export FREERTOS_HEAP_TYPE
+
 ifeq ($(BOARD),Aud01_RevA)
 $(error ERR_NOTSUPPORTED: This project is not supported for the Audio board)
 endif
--- a/Libraries/FreeRTOS/Makefile
+++ b/Libraries/FreeRTOS/Makefile
@@ -67,6 +67,8 @@ else ifeq ($(FREERTOS_HEAP_TYPE), 4)
 SRCS += heap_4.c
 else ifeq ($(FREERTOS_HEAP_TYPE), 5)  
 SRCS += heap_5.c
+else 
+$(error FREERTOS_HEAP_TYPE must be between 1-5. Default: 4)
 endif
 
 # Where to find source files for this project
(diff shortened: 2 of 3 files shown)


Examples/MAX78000/FreeRTOSDemo/project.mk
```diff
@@ -16,6 +16,11 @@ DEBUG=1
 
 LIB_FREERTOS = 1
 
+# Can provide a value for the FREERTOS heap allocation scheme
```
Jake-Carter, line 19: Instead of documenting this here, you can add some info to the UG.

See [here](https://github.com/analogdevicesinc/msdk/blob/main/USERGUIDE.md#sdhc-build-variables) for an example for the SDHC library.

The FreeRTOS lib hasn't had any options before, so I'd suggest adding a similar table

Jake-Carter (changes requested):
Looks good, thanks!  Just one change request for the documentation

Examples/MAX78000/FreeRTOSDemo/project.mk
Brandon-Hurst: Good idea! I'll update once this has been added.

Brandon-Hurst: Should be good now 👍

Jake-Carter (approved):

PR #1080, opened 2024-07-11
fix(BLE): ME18 PHY + PAL UART + clean.cordio

EricB-ADI:
# Description

## ME18 PHY
The ME18 PHY was showing some issues in stability. This build is the most up to date and stable PHY we have.

## PAL UART
``PAL UART`` triggers an assertion by default anytime an error occurs. Now if it is in on the console or the ``HCI UART``, the FIFO is flushed and we continue. Cordio has no mechanism to bubble the error back up, so if an error occurs on HCI, two errors will always occur unless extremely lucky. 
The first error is the actual error from reported by the hardware. The second error is Cordio not being able to recreate an HCI packet since, the packet is incomplete as the previous partial segment is being added to the new segment. 
After this error it starts over and goes back to normal. 

The error always seems to be an overflow error. The best solution for this is to enable HW flow control. 

## clean.cordio
Fixed clean.cordio recipe in libCordio.mk. The ``MAXIM_PATH`` is not properly being passed down, so if you do not have the ``MAXIM_PATH`` set as an environment variable, ``clean.cordio`` does not work. CMSIS_PATH is set so I just used a relative path from that.


kevin-gillespie (approved):

kevin-gillespie:
Not sure if disabling the terminal echo will help with our test automation. I can't tell if the tests now are failing because of the UART in general, the echo, or BLE disconnects.

```diff
diff --git a/Libraries/Cordio/wsf/sources/util/terminal.c b/Libraries/Cordio/wsf/sources/util/terminal.c
index 7d97f548ff..85a8e10ade 100644
--- a/Libraries/Cordio/wsf/sources/util/terminal.c
+++ b/Libraries/Cordio/wsf/sources/util/terminal.c
@@ -101,7 +101,7 @@ void TerminalInit(wsfHandlerId_t handlerId)
   terminalCb.handlerId     = handlerId;
   terminalCb.pFirstCommand = NULL;
   terminalCb.isExecuting   = FALSE;
-  terminalCb.doEcho        = TRUE;
+  terminalCb.doEcho        = FALSE;
   terminalCb.bufOffset     = 0;
 
   TerminalRegisterCommand(&terminalCommandHelp);
```

PR #1081, opened 2024-07-12
feat(Examples): MAX32672 ADC + DMA Example Project

BrentK-ADI:
Add example project utilizing the integrated ADC at fast sample rates with DMA.  Perform RMS calculations on the data and optional convert back to analog with a SPI connect AD5592r.

This project samples a single channel of the integrated ADC at ~500ksps using bank-swapping DMA to provide a continuous sample stream.  RMS is run on the samples, and a moving average filter on the RMS data for smoothing.    The calculated result can be optionally output via the console and/or a SPI connected AD5592r.

This example originated from a customer request but highlighted several features of the MAX32672 that add value as an example project.


BrentK-ADI:
There was a good comment in an offline review regarding using a full citation for the sqrt32 algorithm.  I'll add that to make sure things are attributed correctly.

BrentK-ADI:
@Jake-Carter @lorne-maxim   No rush on merging this. Just a comment to refresh the GitHub stale notification.

Jake-Carter:
@BrentK-ADI thanks for the ping, this got a little buried.  Will review this week

Jake-Carter (approved):

PR #1084, opened 2024-07-22
fix(PeriphDrivers): IBRO_CLK is connected to timers 4-5

isztldav:
According to `data-sheets/MAX78002.pdf` page 43, IBRO is connected to LPTMR0 and LPTMR1.  
I actually tested this as I needed LPTMR0 to output IBRO.  
Therefore, `if (tmr_id > 3)` is wrong.

### DIFF
--- a/Libraries/PeriphDrivers/Source/TMR/tmr_ai87.c
+++ b/Libraries/PeriphDrivers/Source/TMR/tmr_ai87.c
@@ -48,10 +48,6 @@ int MXC_TMR_Init(mxc_tmr_regs_t *tmr, mxc_tmr_cfg_t *cfg, bool init_pins)
         break;
 
     case MXC_TMR_IBRO_CLK:
-        if (tmr_id > 3) { // Timers 4-5 do not support this clock source
-            return E_NOT_SUPPORTED;
-        }
-
         clockSource = MXC_TMR_CLK2;
         MXC_SYS_ClockSourceEnable(MXC_SYS_CLOCK_IBRO);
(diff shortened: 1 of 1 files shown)


Libraries/PeriphDrivers/Source/TMR/tmr_ai87.c
```diff
@@ -48,10 +48,6 @@ int MXC_TMR_Init(mxc_tmr_regs_t *tmr, mxc_tmr_cfg_t *cfg, bool init_pins)
         break;
 
     case MXC_TMR_IBRO_CLK:
-        if (tmr_id > 3) { // Timers 4-5 do not support this clock source
-            return E_NOT_SUPPORTED;
-        }
-
         clockSource = MXC_TMR_CLK2;
```
sihyung-maxim, line 51: The IBRO clock source number is different for LPTMR0-1 vs TMR0-3, so you'll still need to add some logic for that:

```
if (tmr_id <= 3) {
    clockSource = MXC_TMR_CLK2;
} else {
    clockSource = MXC_TMR_CLK0;
}
```

image

sihyung-maxim (changes requested):
It also looks like we need to add support for `IBRO/8` with LPTMR1 (TMR5). Nice catch, thanks for bringing this up.

Libraries/PeriphDrivers/Source/TMR/tmr_ai87.c
isztldav: you are right, fixed

sihyung-maxim (approved):

PR #1091, opened 2024-07-26
feat(Examples): Firmware update through HCI

JC-282-AD:
### Description

Write a Bootloader example: update the firmware through HCI
Add HCI command to be used during the update: erase, update, sysreset


EricB-ADI:
Make sure you merge main in and get rid of merge conflicts.

Examples/MAX32655/Bluetooth/Bootloader/main_int.c
EricB-ADI, line 1: revert changes in the Bootloader example

Libraries/BlePhy/MAX32690/libphy.a
EricB-ADI, line 1: do not commit a new phy. That goes for libphy_hard.a as well

Libraries/Cordio/controller/sources/ble/lhci/lhci_cmd_vs.c
EricB-ADI, line 1: For firmware erase. To be compliant, the HCI has to return a status code. You need to queue an event that resets the device once the task scheduler goes idle. Furthermore, you have to gracefully handle all BLE. So, in this event you need to disconnect if a connection is open shut off the base band and then you can reset.

Libraries/Cordio/controller/sources/common/chci/chci_tr.c
EricB-ADI, line 1: did you modify this?

EricB-ADI (changes requested):
Make sure you merge main first. There is a lot of files change that should not be. Delete any unused files and strip any code that is not used. The only files you should have in this PR are the examples, and the HCI updates inside ``lhci_int.h`` and ``lhci_cmd_vs.c``

PR #1092, opened 2024-07-29
build(CMSIS): pointless RISC-V build system interruption

isztldav:
If you try to build `https://github.com/analogdevicesinc/msdk/tree/main/Examples/MAX78002/RV_ARM_Loader`, this will fail because of a hardcoded interruption. I would propose to remove this line as it does not seem to serve anything and when removed the system builds without issues.


Jake-Carter (approved):
Thanks @isztldav, not sure how this slipped through...

PR #1098, opened 2024-07-30
feat(Examples): Firmware update through HCI for MAX32655

JC-282-AD:
### Description

Write a firmware update example: update the firmware through HCI
Add HCI command to be used during the update: erase, update, sysreset


Libraries/Cordio/controller/sources/ble/lhci/lhci_cmd_vs.c
```diff
@@ -83,10 +92,76 @@ bool_t lhciCommonVsStdDecodeCmdPkt(LhciHdr_t *pHdr, uint8_t *pBuf)
...
+        uint32_t addr;
+        uint32_t size;
+
+
+        BSTREAM_TO_UINT32(addr, pBuf);
+        address = (uint32_t*) addr;
+        uint8_t * addressErase = (uint8_t*) addr;
+
+        BSTREAM_TO_UINT32(size, pBuf);
+        volatile uint32_t address32 = (uint32_t)addressErase;
```
EricB-ADI, line 123: What is addresss32 for? I only see it set

Libraries/Cordio/controller/sources/ble/lhci/lhci_cmd_vs.c
```diff
@@ -64,6 +66,13 @@
 /**************************************************************************************************
   Functions
 **************************************************************************************************/
+//this is the callback function to the Msg
+void device_reset(wsfEventMask_t event, wsfMsgHdr_t *pMsg)
+{
+    MXC_Delay(3000);
```
EricB-ADI, line 72: Do you need the delay?

JC-282-AD: > Do you need the delay?

I think yes, without the delay, the reset will be completed before HCI sends the status code back.

Libraries/Cordio/controller/sources/ble/lhci/lhci_cmd_vs.c
JC-282-AD: > What is addresss32 for? I only see it set

Sorry, this becomes unused after several changes. Deleted.

Libraries/Cordio/controller/sources/ble/lhci/lhci_cmd_vs.c
EricB-ADI: That delay is blocking. meaning nothing is going on when you run it.

EricB-ADI: Instead of doing an event with a delay just use a one 
shot timer

EricB-ADI (approved):

PR #1104, opened 2024-08-02
fix(PeriphDrivers): Fix MAA Operator Setter for TPU Drivers

### LINKED ISSUE #1089: MAX32665 and MAX32666 SDK TPU/MAA Bug
Hello, 

I am working on using the TPU on the MAX32665 and MAX32666 to do hardware acceleration [in wolfSSL](https://github.com/wolfSSL/wolfssl/pull/7777). I found an issue potentially with the process of setting different math modes for the MAA.

I think it has to do with this line:
[#define MXC_SETFIELD(reg, mask, setting) (reg = ((reg) & ~(mask)) | ((setting) & (mask)))](https://github.com/analogdevicesinc/msdk/blob/8fc1336aefd6fcb17d82c1f877bfd8fa5b019d05/Libraries/CMSIS/Device/Maxim/MAX32665/Include/max32665.h#L889C2-L889C8)

I suspect it is not preforming the bit masking correct for the values give for the MAA controls. You can test this by trying your [MAA example](https://github.com/analogdevicesinc/msdk/blob/main/Examples/MAX32665/MAA/main.c) that is in the master branch and switching it to do a multiple operation. You will find it actually does a square operation.

With the current bit masking it will not use values correctly and cut off the last bit of the math control values do not get shifted left by 1 bit so this values will not work as the math operation control register in in bit position [3:1]:
```
        MXC_TPU_MAA_EXP      0b000
        MXC_TPU_MAA_SQ       0b001
        MXC_TPU_MAA_MUL      0b010
        MXC_TPU_MAA_SQMUL    0b011
        MXC_TPU_MAA_ADD      0b100
        MXC_TPU_MAA_SUB      0b101
```

To currently use the MAA I have to define control values as such:
```
        #define WC_MXC_TPU_MAA_EXP      0b0000
        #define WC_MXC_TPU_MAA_SQ       0b0010
        #define WC_MXC_TPU_MAA_MUL      0b0100
        #define WC_MXC_TPU_MAA_SQMUL    0b0110
        #define WC_MXC_TPU_MAA_ADD      0b1000
        #define WC_MXC_TPU_MAA_SUB      0b1010
```
and edit this[ line in the SDK](https://github.com/analogdevicesinc/msdk/blob/8fc1336aefd6fcb17d82c1f877bfd8fa5b019d05/Libraries/PeriphDrivers/Source/TPU/tpu_reva.c#L689):
`if (clc >= 0x6) {`
to:
`if (clc >= 0b1111) {`

To properly use the MAA hardware. This will end up setting the correct bits to use the current bit masking controls.

Jake-Carter:
### Description

Fixes #1089 

### DIFF
--- a/Libraries/PeriphDrivers/Source/TPU/tpu_reva.c
+++ b/Libraries/PeriphDrivers/Source/TPU/tpu_reva.c
@@ -717,7 +717,13 @@ int MXC_TPU_RevA_MAA_Compute(mxc_tpu_reva_regs_t *tpu, mxc_tpu_maa_clcsel_t clc,
     memcpy((void *)MAA_M, (uint32_t *)mod, len);
 
     // Start MAA
-    MXC_SETFIELD(tpu->maa_ctrl, MXC_F_TPU_REVA_MAA_CTRL_CLC, clc << MXC_F_TPU_REVA_MAA_CTRL_CLC);
+    uint8_t clc_val = (uint8_t)clc;
+    if (((tpu->maa_maws & MXC_F_TPU_MAA_MAWS_MSGSZ) >> MXC_F_TPU_MAA_MAWS_MSGSZ_POS) >= 1024) {
+        // The actual value of the field must be left shifted by 1 if MAWS >= 1024
+        clc_val <<= 1;
+    }
+    MXC_SETFIELD(tpu->maa_ctrl, MXC_F_TPU_REVA_MAA_CTRL_CLC,
+                 clc_val << MXC_F_TPU_REVA_MAA_CTRL_CLC_POS);
     tpu->maa_ctrl |= MXC_F_TPU_REVA_MAA_CTRL_STC;
 
     // Check if MAA Error occurs
--- a/Libraries/PeriphDrivers/Source/TPU/tpu_reva.c
+++ b/Libraries/PeriphDrivers/Source/TPU/tpu_reva.c
@@ -261,8 +261,8 @@ int MXC_TPU_RevA_Cipher_Config(mxc_tpu_reva_regs_t *tpu, mxc_tpu_reva_modesel_t 
 // ******************************* Function to Select the Source of the Cipher Key *************************************
 int MXC_TPU_RevA_Cipher_KeySelect(mxc_tpu_reva_regs_t *tpu, mxc_tpu_reva_keysrc_t key_src)
 {
-    MXC_SETFIELD(tpu->cipher_ctrl, MXC_F_TPU_REVA_CIPHER_CTRL_SRC,
-                 key_src << MXC_F_TPU_REVA_CIPHER_CTRL_SRC_POS);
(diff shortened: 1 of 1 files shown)


Jake-Carter:
@night1rider if you can re-test with this fix and confirm functionality we would very much appreciate it.

Libraries/PeriphDrivers/Source/TPU/tpu_reva.c
```diff
@@ -261,7 +261,8 @@ int MXC_TPU_RevA_Cipher_Config(mxc_tpu_reva_regs_t *tpu, mxc_tpu_reva_modesel_t
 // ******************************* Function to Select the Source of the Cipher Key *************************************
 int MXC_TPU_RevA_Cipher_KeySelect(mxc_tpu_reva_regs_t *tpu, mxc_tpu_reva_keysrc_t key_src)
 {
-    MXC_SETFIELD(tpu->cipher_ctrl, MXC_F_TPU_REVA_CIPHER_CTRL_SRC, key_src);
+    MXC_SETFIELD(tpu->cipher_ctrl, MXC_F_TPU_REVA_CIPHER_CTRL_SRC,
+                 key_src << MXC_F_TPU_REVA_CIPHER_CTRL_SRC_POS);
```
ozersa, line 265: [mxc_tpu_reva_keysrc_t](https://github.com/analogdevicesinc/msdk/blob/main/Libraries/PeriphDrivers/Source/TPU/tpu_reva.h#L53) mapped with set values ( _S_ )  instead of value ( _V_ ) so that key_src not need to be shifted.

https://github.com/analogdevicesinc/msdk/blob/main/Libraries/PeriphDrivers/Source/TPU/tpu_reva_regs.h#L213

ozersa (changes requested):
Shifting not require for _keySelect function.

Jake-Carter:
Thanks @night1rider.  What size are you initializing the MAA with?

The reason I ask is that I see some of the TPU control fields are conditional based on the MAA word size (see MAA_MAWS register).  Ex:

image

The field values that you have found to work follow the same scheme for the MAWS > 1024 that some of these other fields do.  I wonder if this is also the case for the clc field but our UG is not correct.

image

Libraries/PeriphDrivers/Source/TPU/tpu_reva.c
Jake-Carter: https://github.com/analogdevicesinc/msdk/pull/1104/commits/851de155d803be14456a7fea99579fdb43c81c99

Jake-Carter:
@night1rider 

(applies palm to face) yep, you're absolutely right...  my mistake, sorry about that typo on that first pass...  `CLC_POS` is a left shift by one and gave the false impression.

Confirmed on the original design spec the values in the UG are correct as well.  Appreciate the testing and patience

Jake-Carter:
@ozersa ready for re-review

ozersa (approved):
Looks good, thanks @Jake-Carter

PR #1127, opened 2024-08-22
docs(Examples): Update SCPA OTP Dump Example README

ozersa:
SCPA examples can not be debugged add it in README file.


Jake-Carter (changes requested):
Thanks @ozersa

The `send_scp` tool is also missing from `Tools/SBT/bin`.  Can you add it as well?

ozersa:
> Thanks @ozersa
> 
> The `send_scp` tool is also missing from `Tools/SBT/bin`. Can you add it as well?

Added it for windows, there is not an executable for linux and macos, For these OSes it is used over python source files.

Examples/MAX32520/SCPA_OTP_Dump/README.md
```diff
@@ -25,7 +25,8 @@ Universal instructions on building, flashing, and debugging this project can be
 
 ### Project-Specific Build Notes
 
-(None - this project builds as a standard example)
+SCP Applet examples are special examples as mentioned in description section, these example can not be debugged as regular example.
+It need to be build with SCPA mode, after build genereated scp packages need to be send ROM Bootlader by SBT/scp_sender tool.
```
sihyung-maxim, line 29: Typos. Maybe something like this?

SCP Applet examples are special examples that can not be debugged like regular examples. They need to be built with SCPA mode. After the build, the generated SCP packages must be sent to the ROM Bootloader using the SBT/scp_sender tool.

sihyung-maxim (dismissed):

Examples/MAX32520/SCPA_OTP_Dump/README.md
ozersa: Thanks, updated.

Jake-Carter:
@ozersa looks like `library.zip` contains source code for a Python interpreter.  I don't see a `send_scp.py` script in there, though.  We may run into licensing issues if we bring the whole interpreter in.

Is there just a `send_scp.py` script we can attach?

ozersa:
> @ozersa looks like `library.zip` contains source code for a Python interpreter. I don't see a `send_scp.py` script in there, though. We may run into licensing issues if we bring the whole interpreter in.
> 
> Is there just a `send_scp.py` script we can attach?

No, there are multiple python file & packages requires to this tool works, please check C:/MaximSDK/Tools/SBT/src/send_scp/src folder.
image

Jake-Carter:
Thanks @ozersa, this makes more sense now since we distribute the full tools through the MSDK installer.

I removed the `library.zip` file and tweaked the readme a little bit more to point to the SBT UG.  (https://github.com/analogdevicesinc/msdk/pull/1127/commits/f721948d7b3dda5a2f08e4701d8a3bb2d669c6f5)

In the future I think the packaging/documentation for the SBTs could be improved.  For now this highlights that the example can't be debugged like usual.

Jake-Carter (approved):

PR #1145, opened 2024-08-29
fix(PeriphDrivers): Flip strong vs weak pull-up bit values

Jake-Carter:
### Description

Apparently most UGs have this flipped.  _Strong_ should set the `ps` field to 0, while _weak_ should set it to `1`.

### DIFF
--- a/Libraries/PeriphDrivers/Source/GPIO/gpio_ai85.c
+++ b/Libraries/PeriphDrivers/Source/GPIO/gpio_ai85.c
@@ -175,28 +175,30 @@ int MXC_GPIO_Config(const mxc_gpio_cfg_t *cfg)
             gpio->padctrl1 &= ~cfg->mask;
             break;
 
+        // Note: for "ps" field set 1 for weak and 0 for strong.
+        // As of 8-28-2024 most UG tables have this flipped the wrong way
         case MXC_GPIO_PAD_WEAK_PULL_UP:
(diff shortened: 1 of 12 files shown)


Jake-Carter:
Pending final confirmation...  don't merge yet

sihyung-maxim (approved):
How was this unnoticed for so long lol

perkinsmg:
There are lots of instances where we use the enums which will now have the opposite effect to what they did before.
Mostly that's probably good, because they were probably not working correctly before.
But in at least the case of the MXC_GPIO_PAD_PULL_UP used in pins_me18.c for the UART signals, Kevin's check-in message said he was adding a weak pull-up, and Lorne recommends a weak pull-up, so I think we should change these to weak to preserve the old behaviour.

Jake-Carter:
> There are lots of instances where we use the enums which will now have the opposite effect to what they did before. Mostly that's probably good, because they were probably not working correctly before. But in at least the case of the MXC_GPIO_PAD_PULL_UP used in pins_me18.c for the UART signals, Kevin's check-in message said he was adding a weak pull-up, and Lorne recommends a weak pull-up, so I think we should change these to weak to preserve the old behaviour.

@perkinsmg updated in https://github.com/analogdevicesinc/msdk/pull/1145/commits/1717e1a721927ac4efddb5189e64570bfbc2b9da

PR #1113, opened 2024-08-13
fix(PeriphDrivers): Add UART Clock Setter API for all parts using UART RevB Drivers

sihyung-maxim:
### Description

This PR is an extended addon to https://github.com/analogdevicesinc/msdk/pull/1108 for the UART RevB drivers. Since the parameters for the `MXC_UART_RevB_SetFrequency` were changed, the parts without these new functions did not properly set the baud rate.

Only tested on ME17 so far.

### Checklist Before Requesting Review

- [ ] PR Title follows correct guidelines.
- [ ] Description of changes and all other relevant information.
- [ ] (Optional) Link any related GitHub issues [using a keyword](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)
- [ ] (Optional) Provide info on any relevant functional testing/validation.  For API changes or significant features, this is not optional.

PR #1150, opened 2024-08-30
fix(PeriphDrivers): Fix UART Functional Regression for MAX32670, MAX32672, MAX32675, MAX32657

Jake-Carter:
### Description

Fixes functional regression caused by #1113 

Calls to MXC_UART_RevB_SetFrequency weren't updated for these parts after changing the RevB API.  As a result, `MXC_UART_RevB_GetTXFIFOAvailable` would eventually hang as the UART FIFO compeltely fills up and is never emptied due to misconfiguration.

Symptoms were that `printfs` would eventually infinitely hang.  i.e. even `Hello_World` wouldn't work.

This PR should fix the regressions.

Also removed `MXC_UART1` and `MXC_UART3` definitions for ME16.  These don't exist on the micro.

### DIFF
--- a/Libraries/CMSIS/Device/Maxim/MAX32675/Include/max32675.h
+++ b/Libraries/CMSIS/Device/Maxim/MAX32675/Include/max32675.h
@@ -438,32 +438,24 @@ typedef enum {
 
 #define MXC_BASE_UART0 ((uint32_t)0x40042000UL)
 #define MXC_UART0 ((mxc_uart_regs_t *)MXC_BASE_UART0)
-#define MXC_BASE_UART1 ((uint32_t)0x40043000UL)
-#define MXC_UART1 ((mxc_uart_regs_t *)MXC_BASE_UART1)
 #define MXC_BASE_UART2 ((uint32_t)0x40044000UL)
 #define MXC_UART2 ((mxc_uart_regs_t *)MXC_BASE_UART2)
-#define MXC_BASE_UART3 ((uint32_t)0x40145000UL)
-#define MXC_UART3 ((mxc_uart_regs_t *)MXC_BASE_UART3)
 
 #define MXC_UART_GET_IRQ(i)             \
     (IRQn_Type)((i) == 0 ? UART0_IRQn : \
-                (i) == 1 ? UART1_IRQn : \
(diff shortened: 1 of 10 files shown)


Libraries/CMSIS/Device/Maxim/MAX32675/Include/max32675.h
```diff
@@ -438,32 +438,16 @@ typedef enum {
...
+
+#define MXC_UART_GET_BASE(i) ((i) == 0 ? MXC_BASE_UART0 : (i) == 2 ? MXC_BASE_UART2 : 0)
 
-#define MXC_UART_GET_UART(i) \
-    ((i) == 0 ? MXC_UART0 : (i) == 1 ? MXC_UART1 : (i) == 2 ? MXC_UART2 : (i) == 3 ? MXC_UART3 : 0)
+#define MXC_UART_GET_UART(i) ((i) == 0 ? MXC_UART0 : (i) == 2 ? MXC_UART2 : 0)
 
-#define MXC_UART_GET_IDX(p) \
-    ((p) == MXC_UART0 ? 0 : (p) == MXC_UART1 ? 1 : (p) == MXC_UART2 ? 2 : (p) == MXC_UART3 ? 3 : -1)
+#define MXC_UART_GET_IDX(p) ((p) == MXC_UART0 ? 0 : (p) == MXC_UART2 ? 2 : -1)
```
sihyung-maxim, line 450: UART1 and UART3 (LPUART0) do exist for the new [ME16 revision](https://www.analog.com/media/en/technical-documentation/data-sheets/max32675c.pdf) (ME16B/C).

sihyung-maxim (changes requested):
Thanks for catching this - To be honest, I only looked at the SVD files to check what parts were using UART RevB. D:

Libraries/CMSIS/Device/Maxim/MAX32675/Include/max32675.h
Jake-Carter: Thanks, reverted and added support for the clocks from table 5 of the datasheet

sihyung-maxim (approved):

PR #1166, opened 2024-09-11
feat(MiscDrivers): Add Ability to Set Pushbutton Polarity

### LINKED ISSUE #1156: ./Libraries/MiscDrivers/PushButton/pb.c PB_Get function doesn't work on AD-APARD32690-SL board
The button on the EvKit1 is set up to be high when open, low when closed, so PB_Get works.
But on AD-APARD32690-SL, it's the other way round, and so PB_Get returns when the button is open, not closed.
Of course, you can fix this at the call-site, but it does mean examples don't port nicely from one board to another.

Jake-Carter:
### Description

Closes #1156 

- Adds the `PB_Set_Polarity` API so that we can adapt the PB drivers to boards with different configurations.
- Initialize the default polarity to `PB_POLARITY_LOW` to preserve backwards compatibility with existing boards.
- Update MAX32690 APARD BSP to set `PB_POLARITY_HIGH` 


sihyung-maxim:
The not enough memory build error for the ME18 Bootloader example has returned, but this time, it's on the ME18APARD board.

perkinsmg (approved):
Looks good, thanks!

sihyung-maxim:
I corrected the Bootloader example's memory configurations to their actual length - which should be enough room to prevent oversize build errors in the future.

For future reference, thee ME18 Bootloader example previously assumed the two flash instances had the same flash page sizes and lengths which is incorrect.

Flash 0: 3MB total, 16KB per page
Flash 1: 256KB total, 8KB per page

PR #1179, opened 2024-09-18
feat(CMSIS): Adding mallinfo function and reworking Cordio memory management.

kevin-gillespie:
### Description

Adding mallinfo function to give us information on how much heap space is being used and how much is available. Reworking the Cordio memory management so we don't have to request an arbitrarily large block of memory on initialization. 


Libraries/Cordio/controller/sources/ble/ll/ll_init.c
```diff
@@ -113,7 +113,7 @@ void LlGetDefaultRunTimeCfg(LlRtCfg_t *pCfg)
 /*************************************************************************************************/
 void LlInitRunTimeCfg(const LlRtCfg_t *pCfg)
 {
-  WSF_ASSERT(pLctrRtCfg == NULL);
+//   WSF_ASSERT(pLctrRtCfg == NULL);
   WSF_ASSERT(pCfg);
```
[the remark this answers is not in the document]
kevin-gillespie: Yes we need to remove this check. The problem is that we will get an assertion if this function is called twice. We need to call this twice to calculate the length of dynamic memory needed.

Libraries/Cordio/wsf/sources/targets/baremetal/wsf_heap.c
```diff
@@ -99,8 +71,9 @@ void WsfHeapAlloc(uint32_t size)
 /*************************************************************************************************/
 void *WsfHeapGetFreeStartAddress(void)
 {
-    if(freeStartAddr == 0) {
-        wsfHeapInit();
+    if(freeStartAddr == (caddr_t)-1) {
+        WSF_ASSERT(0);
+        return NULL;
```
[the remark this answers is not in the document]
kevin-gillespie: Our implementation of sbrk in heap.c returns -1 on failure.

Libraries/Cordio/wsf/sources/targets/baremetal/wsf_heap.c
```diff
@@ -78,16 +58,8 @@ void WsfHeapAlloc(uint32_t size)
...
-        wsfHeapInit();
-    }
-
-    if(freeLen < size) {
-        WSF_ASSERT(FALSE);
-    }
-
-    freeStartAddr += size;
-    freeLen -= size;
+    freeStartAddr = sbrk(size);
```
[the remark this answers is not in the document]
kevin-gillespie: I think the way we have this implemented is correct. We call ```sbrk()``` here to get a memory region of length ```size```. ```sbrk()``` returns heap_end then increments it by ```size```.  ```heap_end``` is inclusive.

First call to ```sbrk()```:
```c
size == 0x40
heap_end == 0
&__HeapBase == 0x2000000
&__HeapLimit == 0x20001000

// heap_end is set to 0x20000000
// previous_heap_end set to 0x20000000
// heap_end incremented to 0x20000040
// return 0x20000000
// calling function can use [0x20000000, 0x2000003F]
```

Second call to ```sbrk()```:
```c
size == 0x80
heap_end == 0x20000040
&__HeapBase == 0x20000000
&__HeapLimit == 0x20001000

// previous_heap_end set to 0x20000040
// heap_end incremented to 0x200000C0
// return 0x20000040
// calling function can use [0x20000040, 0x200000BF]
```

kevin-gillespie:
@crsz20 I was just thinking, it would be good to add a LL function for the size calculation. We're copying a lot of code that could be consolidated.

crsz20:
> @crsz20 I was just thinking, it would be good to add a LL function for the size calculation. We're copying a lot of code that could be consolidated.

Which part do you think could be consolidated? In the `heap.c` files?

kevin-gillespie:
> > @crsz20 I was just thinking, it would be good to add a LL function for the size calculation. We're copying a lot of code that could be consolidated.
> 
> Which part do you think could be consolidated? In the `heap.c` files?

Lines 127 - 133 in the file below. We could add a function with the same parameters as LlInit() that returns the memory requirement. 

https://github.com/analogdevicesinc/msdk/pull/1179/files#diff-7c5e05044eb55068c970128b55bc1b31c278e44f82c0d2325b275d4606af3d9fR127

Libraries/Cordio/controller/sources/ble/init/init.c
```diff
@@ -339,3 +339,24 @@ uint32_t LlInit(LlInitRtCfg_t *pCfg)
...
+/*************************************************************************************************/
+/*!
+ *  \brief  Set LL and BB runtime configurations.
+ *
+ *  \param  pCfg    Runtime configuration.
+ *
+ *  \return Memory used.
+ */
+/*************************************************************************************************/
+uint32_t LlInitMemConfig(LlInitRtCfg_t *pCfg)
```
crsz20, line 352: I decided to name this function *MemConfig* because I see it is doing two things:
- Setting run-time configurations
- Computing the amount of heap memory used

Though, perhaps I could have simply named it `LlInitSetRtCfg()`. I think this is preferable. Let me know what you think.

kevin-gillespie: I like ```LlInitSetRtCfg()```. Take the guts of 322 to 334 and replace those lines with a call to ```LlInitSetRtCfg()```.

crsz20: ### Observation

I noticed in `LlInit()` that the following lines are performed after calling both `LlInitSetBbRtCfg()` and `LlInitSetLlRtCfg()`.

```C
pCfg->pFreeMem += memUsed;
pCfg->freeMemAvail -= memUsed;
```

In the updated examples, on the other hand, these are performed only once after `LlInitSetBbRtCfg()` and omitted after `LlInitSetLlRtCfg()`. Such as, [here](https://github.com/analogdevicesinc/msdk/pull/1179/files#diff-7c5e05044eb55068c970128b55bc1b31c278e44f82c0d2325b275d4606af3d9fR127-R133).

### Conclusion

I am willing to bet that the examples needed that additional increment/decrement of `memUsed`. I will go ahead and add it to our new function `LlInitSetRtCfg()`. Is that correct?

crsz20:
Found a bug in the in BLE4_ctr example on ME17 that seems to date back to the following commit:
- "Reworking HeapAlloc usage" (1a2400e)

The BLE_datc and some other examples seem to run fine, however. I back-traced the error to `PalCfgLoadData()` in main.c > `palCfgLoadBdAddress()` > `MXC_SYS_GetUSN()` failing to an assertion trap in pal_cfg.c.

I am not sure how this may be dependent on the recent changes to the heap memory management.

crsz20:
Seems to be failing at the [checksum verification](https://github.com/analogdevicesinc/msdk/blob/main/Libraries/PeriphDrivers/Source/SYS/sys_me17.c#L129).

kevin-gillespie:
> Seems to be failing at the [checksum verification](https://github.com/analogdevicesinc/msdk/blob/main/Libraries/PeriphDrivers/Source/SYS/sys_me17.c#L129).

This is a common problem with the pre-production parts. Please try a different board to see if we have the same problem.

Libraries/Cordio/controller/sources/ble/init/init.c
kevin-gillespie: I don't think it's significant, but it can't hurt. I'm pretty sure that llCfg structure is only used to pass the initialization paramters. I don't think it's used once it returns from the init function.

yc-adi:
In BLE demos, is it necessary to add a CMD to display the contents of struct mallinfo?

kevin-gillespie:
> In BLE demos, is it necessary to add a CMD to display the contents of struct mallinfo?

No, I don't think so. The memory management is handled behind the scenes.

yc-adi (approved):

PR #1187, opened 2024-09-20
fix(PeriphDrivers): Improve accuracy of timer stop watch

perkinsmg:
### Description

When the timer frequency is small, e.g. 1.8MHz, then dividing this frequency by 1000000 as an integer gives a huge error since it will round down to 1. Improve things by only dividing by 1000 rather than 1000000. We can multiply by 1000 in the numerator, because the prescale value is at most 128, and so 128 * 1000000 fits in uint32_t, and this is then multiplied by ticks which is also 32 bits but widened to 64 bits for the multiply. So the numerator won't overflow.


sihyung-maxim (approved):
Math checks out.

PR #1198, opened 2024-09-30
fix(BLE): Backwards compatibility for HCI Vendor-Specific opcodes

AbbyWolf-ADI:


Libraries/Cordio/controller/sources/ble/lhci/lhci_int.h
```diff
@@ -360,31 +360,29 @@ extern "C" {
 #define LHCI_OPCODE_VS_SET_CONN_PHY_TX_PWR \
     HCI_OPCODE(HCI_OGF_VENDOR_SPEC, 0x3DD) /*!< Set Connection Phy Tx Power opcode. */
 
-#define LHCI_OPCODE_VS_REG_WRITE HCI_OPCODE(HCI_OGF_VENDOR_SPEC, 0x300) /*!< Write data to MCU register or memory space. */
-#define LHCI_OPCODE_VS_REG_READ HCI_OPCODE(HCI_OGF_VENDOR_SPEC, 0x301) /*!< Read data from MCU  register or memory space. */
 #define LHCI_OPCODE_VS_RESET_CONN_STATS \
-    HCI_OPCODE(HCI_OGF_VENDOR_SPEC, 0x302) /*!< Reset connection stats. */
+    HCI_OPCODE(HCI_OGF_VENDOR_SPEC, 0x300) /*!< Reset connection stats. */
```
kevin-gillespie, line 364: Why do we need to change these?

AbbyWolf-ADI: Packetcraft used some of the OCFs that we declared here in for their VS commands in the new stack. I ended up shifting all our VS OCFs down to an unused range, to avoid conflicts like that in the future. For the sake of the HCI Python tools, the VS OCFs used here and the VS OCFs used in the new stack need to match, otherwise we run into compatibility errors.

kevin-gillespie (approved):

PR #1209, opened 2024-10-03
fix(Build): Allow Cordio to Build with Custom BSPs

EricB-ADI:
### Description
Cordio depends on ``led.h`` and ``led.c``, which required ``board.h`` and ``board.c``. If ``LIB_BOARD=0`` cordio fails to build. This is an issue with custom boards not in the MSDK. cordio.mk now uses ``BOARD_DIR`` directly to allow this to override the ``LIB_BOARD`` switch.


Libraries/Cordio/platform/targets/maxim/build/cordio_lib.mk
```diff
@@ -92,11 +92,11 @@ ${CORDIO_BUILD_DIR}/${CORDIO_LIB}: ${CORDIO_C_FILES} ${PROJECTMK}
 	$(MAKE) -f ${CORDIO_DIR}/platform/targets/maxim/build/libCordio.mk lib MAXIM_PATH=${MAXIM_PATH} PROJECT=${CORDIO_LIB} \
 	CORDIO_LIB_VAR=${CORDIO_LIB_VAR} BUILD_DIR=${CORDIO_BUILD_DIR} MFLOAT_ABI=$(MFLOAT_ABI) \
 	DUAL_CORE=$(DUAL_CORE) RISCV_CORE=$(RISCV_CORE) TRACE=${TRACE} DEBUG=${DEBUG} RTOS=${RTOS} \
-	CFG_DEV="${CFG_DEV}" PROJECTMK=${PROJECTMK} BOARD=${BOARD} MXC_OPTIMIZE_CFLAGS=${CORDIO_OPTIMIZE_CFLAGS} \
+	CFG_DEV="${CFG_DEV}" PROJECTMK=${PROJECTMK} BOARD=${BOARD} BSP_SEARCH_DIR=$(BSP_SEARCH_DIR) BOARD_DIR=$(BOARD_DIR) MXC_OPTIMIZE_CFLAGS=${CORDIO_OPTIMIZE_CFLAGS} \
```
Jake-Carter, line 95: @EricB-ADI @kevin-gillespie `LIB_BOARD=0` isn't intended as the mechanism for implementing custom BSPs.  It disables the BSP abstraction entirely (which Cordio is dependent on for the HAL).

Custom BSPs should use the [existing options](https://analogdevicesinc.github.io/msdk//USERGUIDE/#custom-bsps) to provide a "board.h", so I think adding `BSP_SEARCH_DIR` to `CFG_DEV` here would be sufficient to get them working.

Supporting `LIB_BOARD=0` would mean supporting a build without the concept of a BSP which implies no `"board.h"`.  We could check for the `LIB_BOARD` definition to wrap the inclusion.  It gets added here:  https://github.com/analogdevicesinc/msdk/blob/bb1f929917cf228e0a539402ba3046ad164d6d65/Libraries/libs.mk#L35

```diff
diff --git a/Libraries/Cordio/platform/targets/maxim/max32655/sources/pal_uart.c b/Libraries/Cordio/platform/targets/maxim/max32655/sources/pal_uart.c
index 16dd4dd44a..635dfb725d 100644
--- a/Libraries/Cordio/platform/targets/maxim/max32655/sources/pal_uart.c
+++ b/Libraries/Cordio/platform/targets/maxim/max32655/sources/pal_uart.c
@@ -35,7 +35,9 @@
 #include "pal_led.h"
 #include "pal_sys.h"
 
+#ifdef LIB_BOARD
 #include "board.h"
+#endif
 
 #include "uart.h"
 #include "sema.h"
diff --git a/Libraries/MiscDrivers/LED/led.h b/Libraries/MiscDrivers/LED/led.h
index 6ea14a3362..a3353123c9 100644
--- a/Libraries/MiscDrivers/LED/led.h
+++ b/Libraries/MiscDrivers/LED/led.h
@@ -27,7 +27,9 @@
 #define LIBRARIES_MISCDRIVERS_LED_LED_H_
 
 #include "mxc_assert.h"
+#ifdef LIB_BOARD
 #include "board.h"
+#endif
 #include "gpio.h"
 
 #ifdef __cplusplus
```

... but the build error moves to missing definitions on `TERMINAL_UART`, etc.  

`"board.h"` seems the only mechanism for users to pass these in to the HAL, so this may not even be useful.

Jake-Carter (changes requested):

EricB-ADI:
@Jake-Carter I updated it and it seems to be working. You are good to review

Jake-Carter (approved):
Thanks @EricB-ADI 

I caught a couple edge cases.  It seems like the most consistent way to do this is to export `BOARD` and `BSP_SEARCH_DIR`.  exports in general should be used sparingly but in this case I don't see a need to build a library with a different BSP than the app code.

Then, the recursive build call needs to include `board.mk` in its internals.  I re-leveraged `libs.mk` again for this.  In my testing this gets the Cordio lib and app code builds synced up and using the same custom BSP.  It's also more consistent with the direction I'd like to move for libraries in general, like for what we did with TinyUSB (https://github.com/analogdevicesinc/msdk/pull/1101).

PR #1212, opened 2024-10-04
fix(BLE): Add WSF Terminal flush on Console UART error

EricB-ADI:
## Description
- Forcefully reset terminal parser on UART error when used for console

Cordio has no way to report an error upon a hardware failure such as framing or overflow. This will lead to a parser failure. If you are unlucky and get stuck in a place where the new lines are messed up, it may lead to constant bad commands


Libraries/Cordio/platform/targets/maxim/max32655/sources/pal_uart.c
```diff
@@ -92,7 +94,12 @@ void UART_CommonHandler(mxc_uart_regs_t *uart)
   {
     const uint8_t uartIdx = MXC_UART_GET_IDX(uart);
     
-    if( uartIdx == CONSOLE_UART || uartIdx == HCI_UART)
+    if(uartIdx == CONSOLE_UART)
```
kevin-gillespie, line 97: Does this solve the problem we're having? I haven't seen RX errors in the testing that I've done.

EricB-ADI: I have gotten a handful of overflow errors when it is automated.  You probably won't when manually doing it. It used to be an assertion that fired off, which happened 5-10% of runs. It is my best guess right now for the no action bound issue I see.

kevin-gillespie: I think we're better off preventing the overflow with slower writes from the automation.

EricB-ADI: I will close it if you want, but we can't guarantee that the issue wont happen regardless.

PR #1227, opened 2024-10-11
fix(BLE): Converting bool_t to bool

kevin-gillespie:
### Description

Converting bool_t to bool in the types header files.

### DIFF
--- a/Libraries/Cordio/platform/include/pal_types.h
+++ b/Libraries/Cordio/platform/include/pal_types.h
@@ -44,7 +44,11 @@ extern "C" {
 #include <stddef.h>
 
 #ifndef bool_t
-  #define bool_t uint8_t
+  #define bool_t bool
+#endif
+
+#ifndef bool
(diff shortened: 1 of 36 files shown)


Libraries/Cordio/platform/include/pal_types.h
EricB-ADI, line 1: I wouldn't redefine bool. It is in stdbool.h. So just include it under stddef.h.

Also, you may run into an issue that if bool_t is defined, it is not properly defined as bool, you will deal with casting errors like you can if it is uint8_t. 

For TRUE and FALSE, define them as true and false.

Libraries/Cordio/wsf/include/wsf_types.h
EricB-ADI, line 1: I wouldn't redefine everything. Just include pal_types.h

EricB-ADI (approved):

EricB-ADI (approved):

PR #1228, opened 2024-10-11
chore(CMSIS): Generate Registers

EricB-ADI:
### Description
- Updated register docs to include c++ definitions for I/O macros


EricB-ADI:
There is a random svd file that changed? Do not know what it is or where it came from

sihyung-maxim:
> There is a random svd file that changed? Do not know what it is or where it came from

I'll fix the ME13A register files. Right now, the ME13 SVD scripts are set up to generate the ME13B register files in place of the ME13A. This was before it was decided that the ME13B will be renamed from MAX32570 -> MAX32571, so I'll have to separate the two.

sihyung-maxim (approved):

PR #1229, opened 2024-10-13
fix(PeriphDrivers): Fix dma_revb & nvic-table.c warnings

Brandon-Hurst:
RevB DMA driver had incorrect pointer syntax. NVIC driver had an extern inside a function. These generates errors in CI workflows on third-party repositories which use later version of ARM GCC, even though the files aren't necessarily used by a build. In particular, this is blocking a PR merge due to failing CI on Adafruit's CircuitPython. Since the fixes are simple, it makes sense to just get rid of the warnings here. 

https://github.com/adafruit/circuitpython/pull/9667

### Description

- Corrected pointer syntax in dma_revb.c. A couple functions pass srcdst as a pointer instead of a struct. In these cases a '->' instead of a '.' must be used to access members.
- Moved __isr_vector extern outside of NVIC_SetRAM in nvic_table.c. Later versions of GCC create warnings when externs are placed inside of a function. 
- Added "void" argument to a function prototype in dma_revb.h


sihyung-maxim:
I don't know why the DMA RevB drivers are included in the SDK when they're not used anywhere... and especially since no development has been done on them since around ~2020, I'm not surprised there were warnings in them. They don't follow the convention for hardware revision files either.

I think it's safe to delete these drivers entirely.

Brandon-Hurst:
Removed DMA RevB drivers in latest commit

sihyung-maxim (approved):

PR #1236, opened 2024-10-21
fix(BLE): Update PAL UART to derive clock via IBRO

EricB-ADI:
### Description

Previously, the UART clock was fed off of the ``APB`` clock, which is derived from the ``IPO``. Nonproduction parts commonly have inaccurate trims for the ``IPO``. This makes the UART not work as the baud rate is off. ``IBRO`` is a slower more accurate clock, which does not have this issue. 

Changes:
- Initialize all UART using ``pal_uart`` to be on ``IBRO``


EricB-ADI:
MAX32655: 

image

MAX32690:

image

PR #1243, opened 2024-10-25
feat(PeriphDrivers): API to get the package type of part and test date. 

EricB-ADI:
### Description

Starting in 10/23, the USN was updated to include the test date and package type. These APIs cover getting those.


sihyung-maxim (approved):

sihyung-maxim (changes requested):
build not working, unapproving

EricB-ADI:
not stale just busy

Issue #1249, opened 2024-11-01
BLE RF test returns error 48 when performed RX on S2 phy


Ankitpatel2022:
When tried to run the BLE RF test to receive on S2 phy, it returns error 48.
I think the check [here](https://github.com/analogdevicesinc/msdk/blob/main/Libraries/Cordio/controller/sources/ble/ll/ll_main_dtm.c#L870) needs to be modified `((phy < LL_TEST_PHY_LE_1M) || (phy > LL_TEST_PHY_LE_CODED_S2)) ||`

kevin-gillespie:
I understand the problem you're having, but these parameters are in the specification. You can select LL_TEST_PHY_LE_CODED for S2 and S8 when executing the receiver test. The receiver needs to be able to receive both S2 and S8 when using the Coded PHY.
 
![image](https://github.com/user-attachments/assets/f744cf94-5f6c-4ff7-a247-3d965aa8f666)

PR #1253, opened 2024-11-04
fix(BLE): Fixed out of range error in ll_main_dtm.c

EricB-ADI:
### Description

When running an RX test, the ``LlEnhancedRxTest`` Function checked to see if the PHY was within ``1M`` and ``CODED`` which includes ``CODED_S8``, leaving out ``CODED_S2``. This issue does not exist in ``LlEnhancedTxTest``.
The check is updated to make sure it is within ``1M`` and ``CODED_S2``.

Address #1249


EricB-ADI:
This is why S2 tests have shown 100% failure I believe

PR #1259, opened 2024-11-08
feat(Build): Added CXX_STANDARD flag to override default when compiling for C++

EricB-ADI:
### Description
Added CXX_STANDARD flag used in project.mk as ``CXX_STANDARD=14``.Which results in ``-std=c++14`` default value is 17 if not specified


Jake-Carter (approved):
Looks good!

There is an RF_Test build issue

EricB-ADI:
Thanks for the one last approval :)

Jake-Carter:
> Thanks for the one last approval :)

Over and out :)

PR #1267, opened 2024-11-12
fix(BLE): Setting dtmRxSyncMs to 10.

kevin-gillespie:
### Description

This solves a Bluetooth test mode problem when receiving long packets. Increasing the sync time to 10 ms prevents the receiver from timing out while receiving packets from the tester. 


leifangadi (approved):
in DTM heatmap test, increasing this number can help on reducing single-packet loss.  However, it cannot be too long, otherwise coded modes (S2 and S8) at low power region(< -95dBm) will have huge (>30%) PER spikes.

PR #1275, opened 2024-11-19
fix(Examples): Link to the libphy.a directory in MAX32657 BLE5_ctr

crsz20:
### Description

The BLE5_ctr example for the MAX32657 was failing to find the `libphy.a` file. The `project.mk` file was modified accordingly.


EricB-ADI (changes requested):
make this a default like LIB_PHY_DIR ?= ...

EricB-ADI (approved):

PR #1291, opened 2024-12-11
refactor(PeriphDrivers)!: Update WUT API naming schemes for MAX32655, MAX32657, MAX32665, MAX32690, MAX78000, and MAX78002. Offer WUT selection on MAX32690

crsz20:
### Description

The MAX32690's WUT functions perform operations using a single instance of the wake-up timer. However, the [max32690.h](https://github.com/analogdevicesinc/msdk/blob/eeb59b53912d2831e115f18a2a0dad9b5e3f21af/Libraries/CMSIS/Device/Maxim/MAX32690/Include/max32690.h#L369) file shows that there are **two** instances. This can be confirmed in section 24.1 of the device User Guide.

This is a breaking change as the function prototypes and signatures have been modified to offer a `mxc_wut_regs_t *wut` selection as an argument instead of defaulting to a single WUT instance.

### Deprecation

In favor of the `[action][verb]()` naming scheme for functions, the following functions are being marked as deprecated and updated with a new name:

- `MXC_WUT_IntClear()` is being replaced by `MXC_WUT_ClearFlags()`.
- `MXC_WUT_IntStatus()` is being replaced by `MXC_WUT_GetFlags()`.
- `MXC_WUT_Edge()` is being replaced by `MXC_WUT_WaitForEdge()`.
- `MXC_WUT_Store()` is being replaced by `MXC_WUT_StoreCount()`.

**Note**: Unlike the other impacted devices, the MAX32690 will **not** have old functions marked as deprecated. Instead, these will be updated to their new names immediately as part of the breaking change, due to additional function argument. 

### DIFF
--- a/Libraries/PeriphDrivers/Include/MAX32690/wut.h
+++ b/Libraries/PeriphDrivers/Include/MAX32690/wut.h
@@ -158,19 +158,13 @@ uint32_t MXC_WUT_GetCount(mxc_wut_regs_t *wut);
 /**
  * @brief   Clear the timer interrupt.
  */
-__attribute__((deprecated("Use MXC_WUT_ClearFlags instead.  See wut.h for more details."))) void
-MXC_WUT_IntClear(void);
-
-/**
- * @brief   Clear the timer interrupt.
- */
 void MXC_WUT_ClearFlags(mxc_wut_regs_t *wut);
 
 /**
  * @brief   Get the timer interrupt status.
  * @return  Returns the interrupt status. 1 if interrupt has occurred.
  */
-uint32_t MXC_WUT_IntStatus(mxc_wut_regs_t *wut);
+uint32_t MXC_WUT_GetFlags(mxc_wut_regs_t *wut);
 
 /**
  * @brief   Set the timer compare count.
@@ -209,12 +203,12 @@ int MXC_WUT_GetTime(mxc_wut_regs_t *wut, uint32_t ticks, uint32_t *time, mxc_wut
 /**
  * @brief   Wait for an edge of the WUT count register.
  */
-void MXC_WUT_Edge(mxc_wut_regs_t *wut);
+void MXC_WUT_WaitForEdge(mxc_wut_regs_t *wut);
 
 /**
  * @brief   Store the count and snapshot values.
  */
-void MXC_WUT_Store(mxc_wut_regs_t *wut);
+void MXC_WUT_StoreCount(mxc_wut_regs_t *wut);
 
 /**
  * @brief   Restore the DBB clock with the stored count and snapshot values.
--- a/Libraries/PeriphDrivers/Include/MAX32690/wut.h
+++ b/Libraries/PeriphDrivers/Include/MAX32690/wut.h
@@ -158,8 +158,7 @@ uint32_t MXC_WUT_GetCount(mxc_wut_regs_t *wut);
 /**
  * @brief   Clear the timer interrupt.
  */
-__attribute__((
-    deprecated("Use MXC_WUT_ClearFlags instead.  See wut.h for more details.")))void
+__attribute__((deprecated("Use MXC_WUT_ClearFlags instead.  See wut.h for more details."))) void
 MXC_WUT_IntClear(void);
 
 /**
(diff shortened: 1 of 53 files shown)


Libraries/PeriphDrivers/Include/MAX32690/wut.h
```diff
@@ -198,36 +204,36 @@ int MXC_WUT_GetTicks(uint32_t time, mxc_wut_unit_t units, uint32_t *ticks);
...
 /**
  * @brief   Wait for an edge of the WUT count register.
  */
-void MXC_WUT_Edge(void);
+void MXC_WUT_Edge(mxc_wut_regs_t *wut);
 
 /**
  * @brief   Store the count and snapshot values.
  */
-void MXC_WUT_Store(void);
```
crsz20, line 211: Potential alternative name: `MXC_WUT_StoreCount()`

Libraries/PeriphDrivers/Include/MAX32690/wut.h
```diff
@@ -198,36 +204,36 @@ int MXC_WUT_GetTicks(uint32_t time, mxc_wut_unit_t units, uint32_t *ticks);
  * @return     #E_NO_ERROR If everything is successful. 
  * @return     @ref MXC_Error_Codes If function is unsuccessful.
  */
-int MXC_WUT_GetTime(uint32_t ticks, uint32_t *time, mxc_wut_unit_t *units);
+int MXC_WUT_GetTime(mxc_wut_regs_t *wut, uint32_t ticks, uint32_t *time, mxc_wut_unit_t *units);
 
 /**
  * @brief   Wait for an edge of the WUT count register.
  */
-void MXC_WUT_Edge(void);
```
crsz20, line 206: Potential alternative name: `MXC_WUT_WaitForEdge()`

sihyung-maxim: I think that's a good idea. It's a better self-descriptive function name than just `Edge`

Libraries/PeriphDrivers/Include/MAX32690/wut.h
```diff
@@ -114,71 +114,77 @@ typedef void (*mxc_wut_complete_cb_t)(int result);
...
+ * @brief   Clear the timer interrupt.
+ */
+void MXC_WUT_ClearFlags(mxc_wut_regs_t *wut);
 
 /**
  * @brief   Get the timer interrupt status.
  * @return  Returns the interrupt status. 1 if interrupt has occurred.
  */
-uint32_t MXC_WUT_IntStatus(void);
+uint32_t MXC_WUT_IntStatus(mxc_wut_regs_t *wut);
```
sihyung-maxim, line 173: `MXC_WUT_IntStatus(...)` should be replaced with `MXC_WUT_GetFlags(...)` since that's using the old names.

Libraries/PeriphDrivers/Include/MAX32690/wut.h
```diff
@@ -114,71 +114,77 @@ typedef void (*mxc_wut_complete_cb_t)(int result);
...
  */
-uint32_t MXC_WUT_GetCount(void);
+uint32_t MXC_WUT_GetCount(mxc_wut_regs_t *wut);
 
 /**
  * @brief   Clear the timer interrupt.
  */
-void MXC_WUT_IntClear(void);
+__attribute__((deprecated("Use MXC_WUT_ClearFlags instead.  See wut.h for more details."))) void
+MXC_WUT_IntClear(void);
```
sihyung-maxim, line 162: Because many of these changes are breaking changes, I wonder if the deprecated attribute is even worth adding now...

sihyung-maxim (changes requested):
Core changes look good.

Libraries/PeriphDrivers/Include/MAX32690/wut.h
crsz20: I think it makes sense to remove the `MXC_WUT_IntClear()` and deprecation attributes. In practice, I can imagine it would be best for library users to be required to update their application code only once. This function would also be impacted by the WUT selection argument supported with `ClearFlags()`, anyways.

crsz20: Update: Will remove the deprecated attribute just for ME18 due to the breaking change. Other devices will keep the deprecated attribute since only the function names are changing.

crsz20:
To do: Update ME18's wut.h Doxygen comments

crsz20:
The PR description has been updated to account for the ME18 _not_ having functions marked as deprecated.

sihyung-maxim (approved):
Looks good!

PR #1297, opened 2024-12-13
feat(CMSIS,PeriphDrivers): Enable external clock selection on MAX32657

ttmut:
### Description

Uncomment clock selection code to allow external clock input to be used as system clock since external clock pin and system clock selection bits are now known.

Set INRO clock frequency to 131kHz.

### DIFF
--- a/Libraries/CMSIS/Device/Maxim/MAX32657/Include/system_max32657.h
+++ b/Libraries/CMSIS/Device/Maxim/MAX32657/Include/system_max32657.h
@@ -38,7 +38,7 @@ extern "C" {
          Update if use of this oscillator requires precise timing.*/
 /* NOTE: INRO was previously named NANORING */
 #ifndef INRO_FREQ
-#define INRO_FREQ 100000
+#define INRO_FREQ 131000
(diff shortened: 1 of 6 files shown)


Libraries/CMSIS/Device/Maxim/MAX32657/Include/system_max32657.h
```diff
@@ -38,7 +38,7 @@ extern "C" {
          Update if use of this oscillator requires precise timing.*/
 /* NOTE: INRO was previously named NANORING */
 #ifndef INRO_FREQ
-#define INRO_FREQ 100000
+#define INRO_FREQ 131000
```
sihyung-maxim, line 41: Is this the nominal value that's indicated in the datasheet? The INRO frequency is highly variable which could be 131KHz on one part or 80KHz on another.

ttmut: Yes, it is given as 131kHz in both user guide and datasheet for MAX32657, if that is what you mean.

PR #1300, opened 2024-12-16
fix(PeriphDrivers): Fix flash ecc issue for MAX32657

mertvatansever:
### Description

Erasing flash will also erase the ECC bits. These bits are not updated until a flash write. Reading from erased memory will signal a ECC error that is falsely corrected from 0xFF to 0xFD on the 16th byte of each 128-bit line.

This PR applies a workaround by setting the 16th byte of each line to 0xFF to handle this issue for max32657.

This is tested with zephyr/tests/drivers/flash/common. The result is indicated below.

image


ttmut (commented):
What if first 15 bytes are not all 0xFF?

mertvatansever:
> What if first 15 bytes are not all 0xFF?

Updated the code to set 16th byte to 0xFF if 16th byte is 0xFD and rest of the bytes are 0xFF.

kevin-gillespie (approved):

gkhnclkmxm (approved):

ttmut (approved):

MaureenHelm (approved):

PR #1319, opened 2025-01-09
fix(MAXUSB): Updated makefiles for USB examples 

EricB-ADI:
### Description
- Corrected depth used for relative paths when MAXIM_PATH is not set


crsz20 (approved):
Nice catch

PR #1306, opened 2024-12-18
feat(Other): Add zephyr CAN wrapper for MAX32662, MAX32690

trupples:
### Description

The MSDK API for `MXC_CAN_Init` differs between the MAX32662 and MAX32690 boards. The former expects an additional GPIO configuration parameter. In Zephyr, because GPIO config is done by the zephyr layer and not msdk, this parameter is unused anyways (see `MSDK_NO_GPIO_CLK_INIT`).

This PR adds a `Wrap_MXC_CAN_Init` function to allow the `can_max32` driver (currently only available via innersource) to work on the MAX32662 as well (it currently only supports MAX32690). This seems to be the only tweak required to support this board.

I have tested that this properly works with a basic CAN send-recv example on a MAX32662 board and that it builds successfully for the MAX32690.

### Checklist Before Requesting Review

- [x] PR Title follows correct guidelines.
- [x] Description of changes and all other relevant information.

PR #1320, opened 2025-01-09
fix(Other): Fix Wrap_MXC_CAN_Init signature

trupples:
### Description

Fixes a bug introduced by #1306, which added a Zephyr wrapper for `MXC_CAN_Init`, whose signature differs between the MAX32662 and MAX32690.

The previous implementation declared `Wrap_MXC_CAN_Init` as `void` and discarded the return value of the underlying `MXC_CAN_Init` calls. This is incompatible with the can_max32 driver which does error checking and expects an `int` return value.

While this is a "breaking" change, no code currently depends on this wrapper, because its signature is invalid and could not be properly used in the can_max32 driver.

#### How did this happen?

This is entirely my fault, as I was rushing to get some zephyr can_max32 changes in, on the last Friday before the winter holiday break. (I am now well aware of how stupid that is...)

While working on this wrapper, I used the following forks: [trupples/msdk](https://github.com/trupples/msdk), [trupples/hal_adi](https://github.com/trupples/hal_adi). I *did* catch this error and [fixed it in hal_adi](https://github.com/trupples/hal_adi/commit/55ebe41bc769ab4532e3e2cf7a7676cd73f5c5ea), but **didn't push the same changes to msdk**. 

At that point, trupples/hal_adi had the correct implementation (int return), but trupples/msdk did not (void return).

I was able to build and test an application using the driver, because in zephyr I only referenced my hal_adi repo. In PR #1306 **I claimed the wrapper is correct and tested**, and the PR got accepted, but with the wrong code. The wrong code was then automatically copied to [analogdevicesinc/hal_adi](https://github.com/analogdevicesinc/hal_adi).

NB: should have listened to [shouldideploy.today](https://shouldideploy.today/)

#### Validation

A minimal example to test this on both MAX32662 and MAX32690 can be found at [trupples/max32-can-validate](https://github.com/trupples/max32-can-validate/). It unfortunately requires some fiddling around to get the WIP driver from innersource. Most importantly, the entire process is automated by [`validate.sh`](https://github.com/trupples/max32-can-validate/blob/main/validate.sh), which:

1. Clones hal_adi (develop branch), msdk (the branch in this PR)
2. Uses zephyr-hal.sh to properly copy files to hal_adi
3. Prepares the west workspace
4. Overwrites hal_adi with modified variant (this is the step I'm most doubtful of, because it intentionally bypasses `west update` to more easily apply the changes)
5. Builds example for MAX32690 and MAX32662

Usage:
```sh
cd /path/to/west_workspace
git clone https://github.com/trupples/max32-can-validate/ validate
cd validate
./validate.sh
```

I have successfully run this, and I am quite confident that I am not wasting your time with a second bad PR, but I'd like a sanity check.

### DIFF
--- a/Libraries/zephyr/MAX/Include/wrap_max32_can.h
+++ b/Libraries/zephyr/MAX/Include/wrap_max32_can.h
@@ -1,6 +1,6 @@
 /******************************************************************************
  *
- * Copyright (C) 2024 Analog Devices, Inc.
+ * Copyright (C) 2024-2025 Analog Devices, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -26,9 +26,9 @@
 extern "C" {
 #endif
 
-static inline void Wrap_MXC_CAN_Init(uint32_t can_idx, mxc_can_obj_cfg_t cfg,
-                                     mxc_can_unit_event_cb_t unit_cb,
-                                     mxc_can_object_event_cb_t obj_cb)
+static inline int Wrap_MXC_CAN_Init(uint32_t can_idx, mxc_can_obj_cfg_t cfg,
+                                    mxc_can_unit_event_cb_t unit_cb,
+                                    mxc_can_object_event_cb_t obj_cb)
 {
     /* The only API difference between the MAX32662 and MAX32690 relevant to
      * implementing the max32 CAN driver in Zephyr is that the former requires a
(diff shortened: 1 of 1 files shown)


Libraries/zephyr/MAX/Include/wrap_max32_can.h
```diff
@@ -1,6 +1,6 @@
 /******************************************************************************
  *
- * Copyright (C) 2024 Analog Devices, Inc.
+ * Copyright (C) 2025 Analog Devices, Inc.
```
sihyung-maxim, line 3: `Copyright (C) 2024-2025 Analog Devices, Inc.`

sihyung-maxim (changes requested):
No worries, thanks for catching this. Just an unfortunate consequence that we don't have workflows to check for everything.

sihyung-maxim:
Interesting that the Build_Examples workflow is having trouble with your fork. I'm in the process updating this workflow at the moment, but merging this PR because the Zephyr files are not relevant with the Build_Examples workflow.

PR #1330, opened 2025-01-21
feat(Examples): Expand MAX32650 SPI example to select SPI instance

BrentK-ADI:
Updated the MAX32650 SPI example project to allow the user to select the SPI instance to test. All of the driver logic remained the same, the additions updated the instance and IRQ constants and loopback instructions during startup, as well as documentation.

This was based on a customer request regarding confusion on the SPI drivers, as well as identifying the MISO and MOSI lines when leveraging SPI3.


sihyung-maxim (approved):
Thanks Brent.

PR #1334, opened 2025-02-04
fix(PeriphDrivers): Fix UART RevB Driver not cleaning up Async Transactions

Brandon-Hurst:
## Description
Currently, the private AsyncTxRequests and AsyncRxRequests arrays are used to track transactions in progress for uart_revb.c. The Tx/Rx Async Callbacks check for a callback function before clearing the Requests:

```C
mxc_uart_req_t *req = (mxc_uart_req_t *)AsyncTxRequests[uart_num];
if ((req != NULL) && (req->callback != NULL)) {
    AsyncTxRequests[uart_num] = NULL;
    req->callback(req, retVal);
}
```

This means that if the user does not supply an explicit callback function, a transaction never gets cleared, causing problems after just one transaction.

This commit resolves this by moving the cleanup of this important struct outside of the conditional check for a callback function.

```C
mxc_uart_req_t *req = (mxc_uart_req_t *)AsyncTxRequests[uart_num];
if ((req != NULL) && (req->callback != NULL)) {
    req->callback(req, retVal);
}

// Cleanup Async Transaction
AsyncTxRequests[uart_num] = NULL;
```

## Testing 
This was tested as part of development for UART functionality in the CircuitPython BUSIO.UART module:
https://github.com/Brandon-Hurst/circuitpython/blob/ports/analog/add-busio/ports/analog/common-hal/busio/UART.c

I found that it was necessary to submit a callback function for the Async API to work correctly. This fix is to correct this, so the user doesn't have to supply a callback to have the Async Requests cleaned up. 

### DIFF
--- a/Libraries/PeriphDrivers/Source/UART/uart_revb.c
+++ b/Libraries/PeriphDrivers/Source/UART/uart_revb.c
@@ -749,9 +749,11 @@ int MXC_UART_RevB_AsyncTxCallback(mxc_uart_revb_regs_t *uart, int retVal)
 
     mxc_uart_req_t *req = (mxc_uart_req_t *)AsyncTxRequests[uart_num];
     if ((req != NULL) && (req->callback != NULL)) {
-        AsyncTxRequests[uart_num] = NULL;
         req->callback(req, retVal);
     }
+
+    // Cleanup Async Transaction
+    AsyncTxRequests[uart_num] = NULL;
 
     return E_NO_ERROR;
 }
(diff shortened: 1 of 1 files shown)


Brandon-Hurst:
This PR is going stale; commenting to refresh it. Please consider reviewing this change for acceptance when it is convenient.

ttmut (approved):

sihyung-maxim (approved):

PR #1344, opened 2025-02-10
fix(Other): Zephyr: MAX32650: Add missing LP wrappers

Burakbabao:
This commit adds the missing Low Power defines for MAX32650 for the zephyr build system.

There are generally two main groups in the wrap_max32_xx files. MAX32650 fits the first group for Low Power. However, there is no MXC_LP_EnterShutDownMode() function provided for MAX32650 so the MXC_LP_EnterBackupMode() function was chosen as best fit.

### DIFF
--- a/Libraries/zephyr/MAX/Include/wrap_max32_lp.h
+++ b/Libraries/zephyr/MAX/Include/wrap_max32_lp.h
@@ -31,7 +31,7 @@ extern "C" {
  */
 #if defined(CONFIG_SOC_MAX32665) || defined(CONFIG_SOC_MAX32666) || \
     defined(CONFIG_SOC_MAX32670) || defined(CONFIG_SOC_MAX32672) || \
-    defined(CONFIG_SOC_MAX32662) || defined(CONFIG_SOC_MAX32675)
+    defined(CONFIG_SOC_MAX32662) || defined(CONFIG_SOC_MAX32675) || defined(CONFIG_SOC_MAX32650)
 
 static inline void Wrap_MXC_LP_EnterLowPowerMode(void)
 {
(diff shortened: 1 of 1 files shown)


Libraries/zephyr/MAX/Include/wrap_max32_lp.h
```diff
@@ -50,7 +51,11 @@ static inline void Wrap_MXC_LP_EnterStandbyMode(void)
 
 static inline void Wrap_MXC_LP_EnterPowerDownMode(void)
 {
+#if defined(CONFIG_SOC_MAX32650)
+    MXC_LP_EnterBackupMode();
```
hfakkiz, line 55: I think this is not correct mode for this function. According to user guide, there is "Shutdown Mode" for MAX32650 too.
image

Also, there are macros for this mode.
image


Maybe you should create a new "MXC_LP_EnterShutdownMode(...)" function in lp_me10.c file.

Burakbabao: Mb, didn't realize we could touch those as well. Was trying the best fit with what was provided while only touching the wrapper files. Looking into it.

Burakbabao: Done.

ttmut:
Have you tested this on a board?

Burakbabao:
> Have you tested this on a board?

Working on it, there seems to be an issue

Burakbabao:
> Have you tested this on a board?

Soo, the shutdown function bricked two boards. 

@hfakkiz although there is a shutdown mode and the relevant GCR info in the documentation, there seems to be an issue with it. So, for now, the backup function was used.

The issue may be found if the conflicting declarations of the shutdown mode bits are investigated more:

sdown_b011_page67_userguide32650

sdown_b111_page53_userguide32650

ttmut (approved):

PR #1347, opened 2025-02-13
fix(PeriphDrivers, Other): Updates about getting USN

hfakkiz:
### Description

There is a function mismatch for `MXC_SYS_GetUSN(...)` function between MAX32650 and others. To handle these differences, I defined wrapper version of this function. 

Also, I removed magic number for USN length and added related definition for it. I noticed `MXC_SYS_GetUSN(...)` is not called anywhere for MAX32650 so I removed `len` input from function to simplify it and make it similar to other functions.


Libraries/PeriphDrivers/Include/MAX32650/mxc_sys.h
```diff
@@ -416,10 +417,9 @@ uint8_t MXC_SYS_GetRev(void);
 /**
  * @brief      Get the USN of the chip
  * @param      serialNumber buffer to store the USN
- * @param      len          length of the USN buffer
  * @returns    #E_NO_ERROR if everything is successful.
  */
-int MXC_SYS_GetUSN(uint8_t *serialNumber, int len);
+int MXC_SYS_GetUSN(uint8_t *serialNumber);
```
ttmut, line 422: I am conflicted about this. There might be downstream users using the old definition. This change will break their builds. On the other hand, other SoCs do not use the `len` variable so it makes sense to match the definitions.

hfakkiz: Removed this change.

yasinustunerg:
Max32660 is waiting for this PR

ttmut (approved):

PR #1354, opened 2025-02-20
fix(PeriphDrivers): MAX32650 (ME10) Flash Address Fix

BrentK-ADI:
Corrected a bug in how the flash physical address (relative to flash base) was calculated which caused a block of pages' addresses to be incorrectly masked. This ultimately caused issues when doing single page erases whenever bit 20 of the Flash address was set.

In the original implementation, -1 was done on the FLASH_MEM_SIZE to presumably get a valid mask. This works when the memory size is a power of 2.  However, since the memory on the MAX32650 is 3MB (0x300000), subtracting 1 results in a mask of 0x2FFFFF, where Bit 20 is 0.  All pages with Bit 20 set to 1 will have an incorrect address calculated.

The solution was to just subtract the base address, since the range was already validated in the if block.  This is the same implementation that the MAX32690 (also with 3MB of Flash) does.

### DIFF
--- a/Libraries/PeriphDrivers/Source/FLC/flc_me10.c
+++ b/Libraries/PeriphDrivers/Source/FLC/flc_me10.c
@@ -5,9 +5,9 @@
  */
 /******************************************************************************
  *
- * Copyright (C) 2022-2023 Maxim Integrated Products, Inc. (now owned by 
+ * Copyright (C) 2022-2023 Maxim Integrated Products, Inc. (now owned by
  * Analog Devices, Inc.),
- * Copyright (C) 2023-2024 Analog Devices, Inc.
+ * Copyright (C) 2023-2025 Analog Devices, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


Libraries/PeriphDrivers/Source/FLC/flc_me10.c
```diff
@@ -5,7 +5,7 @@
  */
 /******************************************************************************
  *
- * Copyright (C) 2022-2023 Maxim Integrated Products, Inc. (now owned by 
+ * Copyright (C) 2022-2023 Maxim Integrated Products, Inc. (now owned by
  * Analog Devices, Inc.),
  * Copyright (C) 2023-2024 Analog Devices, Inc.
```
Burakbabao, line 10: Should be 2023-2025

BrentK-ADI:
Sorry for the erroneous push. Committed a different fix to the wrong branch on my end. This PR should be back to just the flash fix.

BrentK-ADI:
@Burakbabao Received a Github stale PR notification. Requesting review and merge.

hfakkiz (approved):

Burakbabao (approved):

Issue #1293, opened 2024-12-12
MAX32670 I2C Driver causes a Double-Read


Brandon-Hurst:
Reading from a peripheral, it appears that the I2C driver for MAX32670 causes a double read. 

## What I Did

Connect an ADXL345 to I2C1 on MAX32670 EVKit. I am running the I2C example, and simply added the below source code before the main test runs...

```C
    // FIXME: For testing
    uint8_t my_adxl_tx[1] = {0x0}; // ADXL ID Register
    uint8_t my_adxl_rx[2] = {0x0};
    mxc_i2c_req_t reqMaster = {
        .i2c = I2C_MASTER,
        .addr = 0x1D, // ADXL345 I2C Addr
        .tx_buf = my_adxl_tx,
        .rx_buf = my_adxl_rx,
        .tx_len = 1,
        .rx_len = 2,
        .callback = NULL,
    };
    I2C_FLAG=1;
    MXC_I2C_MasterTransaction(&reqMaster);
``` 

## Outcome

![Double read of ID Register ](https://github.com/user-attachments/assets/5b70e2c8-4688-4220-bbb6-e1f485e8a61d)

The above shows a repeated start and a double read of the register despite only attempting to read once. 

I am trying to use MAX32670 with a No-OS Project that I built and have actually seen this condition cause infinite loops where the I2C bus sends out clocks forever and waits for data that will not come.

## More info 

I dug into the i2c_reva.c driver and it looks like somehow the I2C registers are getting corrupted between when the data comes into the FIFO and when it is actually read. 

Here is when the data has arrived in the FIFO:
![Data in FIFO](https://github.com/user-attachments/assets/d1aebfea-abe5-4a18-afbd-d758cb57aed5)

Here's what happens if I single-step a couple assembly instruction past that point:
![FIFO Empties](https://github.com/user-attachments/assets/67192cf2-c986-4fe5-b72e-1272b2126a88)
![CPU Registers](https://github.com/user-attachments/assets/168d4502-a522-4aa4-aba1-c100347eceba)

There are no interrupts / DMA enabled, and the data in CPU registers r2/r6 are just some values from RAM. 

This causes the `i2c_reva.c` MasterTransaction function to set the Restart bit and attempt to read again, when no more data will come back from the peripheral.

Brandon-Hurst:
May be related to https://github.com/analogdevicesinc/msdk/pull/1211. It looks like the behavior I'm seeing is similar.

[...]

PR #1359, opened 2025-03-05
fix(PeriphDrivers): Issue #1293: MAX32670 I2C Driver causes a Double-Read

dnguye14-adi:
### Description

In the Master Transaction function, add a checking to return BUSY if the I2C bus is not ready.

### DIFF
--- a/Libraries/PeriphDrivers/Source/I2C/i2c_reva.c
+++ b/Libraries/PeriphDrivers/Source/I2C/i2c_reva.c
@@ -902,6 +902,11 @@ int MXC_I2C_RevA_MasterTransaction(mxc_i2c_reva_req_t *req)
         return E_BAD_STATE;
     }
 
+    // Check and return BUSY if the Bus is not ready
+    if (i2c->status & MXC_F_I2C_REVA_STATUS_BUSY) {
+        return E_BUSY;
+    }
+
     // if(!read | write)
     //  Start
     //  send addr w/ write bit


Brandon-Hurst:
Hello, is there any response to my request above or info on when this change could be merged?

sihyung-maxim:
> Hello, as I mentioned in the issue #1293, I recommend changing the first while loop to an if statement. This will prevent the possibility of infinite loops if the I2C peripheral is busy / hanging. Returning an error indicating the I2C controller is busy is sufficient in this case, and a while loop should not be needed. I tested this as well and left screenshots in the issue linked.
> 
> I would propose changing this:
> 
> ```c
> while(i2c->status & MXC_F_I2C_REVA_STATUS_BUSY) {}
> // Wait until last transaction finished and bus ready
> ```
> 
> to this:
> 
> ```c
> // If there is an active transaction, return a BUSY error
> if (i2c->status & MXC_F_I2C_REVA_STATUS_BUSY) {
>     return E_BUSY;
> }
> ```

Sorry, I've been away from MSDK support for a while because of other pending tasks. I agree with this too; we should avoid an infinite loop without a timeout or any way to exit.

AlbertFrancis07:
> Hello, as I mentioned in the issue #1293, I recommend changing the first while loop to an if statement. This will prevent the possibility of infinite loops if the I2C peripheral is busy / hanging. Returning an error indicating the I2C controller is busy is sufficient in this case, and a while loop should not be needed. I tested this as well and left screenshots in the issue linked.
> 
> I would propose changing this:
> 
> ```c
> while(i2c->status & MXC_F_I2C_REVA_STATUS_BUSY) {}
> // Wait until last transaction finished and bus ready
> ```
> 
> to this:
> 
> ```c
> // If there is an active transaction, return a BUSY error
> if (i2c->status & MXC_F_I2C_REVA_STATUS_BUSY) {
>     return E_BUSY;
> }
> ```

I too faced the same Problem of I2C hanging in infinite loop  due to sudden transaction ,  while interfacing MAX78000 with ADXL382 accelerometer and the above proposed solution of making I2C bus  wait  till end of trasaction has helped me fix it.

dnguye14-adi:
@AlbertFrancis07, you could also try with the solution not using the while-loop.

ttmut (approved):

PR #1360, opened 2025-03-05
feat(CMSIS): CFSIO-4121 Call PeripheralInit

emmet-horgan:
### Description
The motivation behind this PR is tool allow configuration tools to generate a `PeripheralInit` function which will automatically be called before `main` so that the user does not have to call it directly in their application.

This PR adds a weakly defined function `PeripheralInit` for each SoC which is called in `SystemInit` after `Board_Init` to initialize the peripherals

The call to `PeripheralInit` is intentionally after `Board_Init` so that the user defined peripheral initialization is not overwritten. I would also question whether or not `PinInit` should be called after `Board_Init` also ? 

### DIFF
--- a/Libraries/CMSIS/Device/Maxim/MAX32572/Source/system_riscv_max32572.c
+++ b/Libraries/CMSIS/Device/Maxim/MAX32572/Source/system_riscv_max32572.c
@@ -126,7 +126,7 @@ __weak void SystemInit(void)
     __enable_irq();
 
     Board_Init();
-    
+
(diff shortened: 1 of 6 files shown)


perkinsmg:
We're going to need the same thing in the RISC-V variants of SystemInit - e.g. ./Libraries/CMSIS/Device/Maxim/MAX32690/Source/system_riscv_max32690.c

lorne-maxim (approved):

PR #1361, opened 2025-03-07
feat(Other): Changes in preparation for Zephyr HPB support

petejohanson-adi:
### Description

The following two commits are prep work for the Zephyr HPB memory-controller driver. They are tested and confirmed working with that driver.

The `const` flags are a nicety, to avoid compiler warnings related to passing in our const structures that are generated from the device tree there.

### DIFF
--- a/Libraries/PeriphDrivers/Include/MAX32650/hpb.h
+++ b/Libraries/PeriphDrivers/Include/MAX32650/hpb.h
@@ -124,7 +124,7 @@ typedef struct {
     mxc_hpb_device_t device_type;
 
     /** Pointer to array of address offset/value pairs */
-    mxc_hpb_cfg_reg_val_t *cfg_reg_val;
+    const mxc_hpb_cfg_reg_val_t *cfg_reg_val;
 
     /** number of configuration pairs */
     unsigned int cfg_reg_val_len;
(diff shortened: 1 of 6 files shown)


Libraries/zephyr/MAX/Include/wrap_max32_emcc.h
```diff
@@ -0,0 +1,43 @@
...
+
+static inline void Wrap_MXC_EMCC_Enable(void)
+{
+    MXC_EMCC_Enable();
+}
+
+static inline void Wrap_MXC_EMCC_Disable(void)
+{
+    MXC_EMCC_Disable();
+}
```
ttmut, line 37: You can just call `MXC_EMCC_Enable()` from inside the Zephyr driver as long as the function prototype is the same for all parts. Wrapping is not needed in that case.

petejohanson-adi: Makes sense, thanks! Removed that commit from the branch.

ttmut: It looks like the same can be applied to HPB wrapper as well.

petejohanson-adi: Indeed. Rebased and removed the commit, and updated my Zephyr driver work to make the MSDK call directly. Thanks.

petejohanson-adi:
@ttmut It *seems* that Ci does not like my PR to include more than one self-contained commit, based on my quick scan of the failure at https://github.com/analogdevicesinc/msdk/actions/runs/13792457914/job/38575624307?pr=1361

Would you rather I:

* Split into two PRs, OR
* Squash into one commit, OR
* Leave it.

TIA.

petejohanson-adi:
> 
> > Would you rather I:
> > 
> > * Split into two PRs, OR
> > * Squash into one commit, OR
> > * Leave it.
> > 
> > TIA.
> 
> Split into two PRs

Split into #1362 and #1363 and rebased both on latest `main`. CI seems to get past the "Get Changed Files" step, so closing this PR in favor of those two discrete PRs.

PR #1377, opened 2025-04-14
fix(PeriphDrivers): MSDK-1287: Fix MAX32655 MXC_Delay on RISC-V shows different behavior when not compiling with -O0

dnguye14-adi:
### Description

In the MSDK Dual Core Sync example for MAX32655, the LED was programed to turn on in 500ms and then turn off.
Issue: when compiling with -O0, the LED stayed on for 500ms correctly. But, when compiling with -O1, -O2, -Os and -Og, the LED stayed on for longer (around 33% longer than the expected one).
To reproduce, just compile the example changing the MXC_OPTIMIZE_CFLAGS accordingly.

Fix: Add a delay to prevent the CPU bothers PCCR too much, then the PCCR hardware can maintain the correct timing.

_______________________________________________________________________________

### DIFF
--- a/Libraries/PeriphDrivers/Source/SYS/mxc_delay.c
+++ b/Libraries/PeriphDrivers/Source/SYS/mxc_delay.c
@@ -42,8 +42,18 @@ int MXC_Delay(uint32_t us)
     CSR_SetPCER(1); // Enable counting of cycles
     CSR_SetPCMR(3); // Turn on counter
 
+    uint32_t num_ticks_half_us = MXC_SYS_RiscVClockRate() / 1000000 / 3;
+    volatile uint32_t counter;
+
     while (CSR_GetPCCR() < ticks) {
-        // Wait for counter to reach the tick count.
+        // Wait about 1 us
+        // and re-check if the counter reaches the tick count or not.
+        // Need this waiting to prevent CPU bothers PCCR too much
+        // to avoid PCCR hardware cannot maintain the correct timing.
+        counter = 0;
+        while (counter < num_ticks_half_us) {
+            counter++;
+        }
     }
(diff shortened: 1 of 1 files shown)


Libraries/PeriphDrivers/Source/SYS/mxc_delay.c
```diff
@@ -42,8 +42,18 @@ int MXC_Delay(uint32_t us)
     CSR_SetPCER(1); // Enable counting of cycles
     CSR_SetPCMR(3); // Turn on counter
 
+    uint32_t num_ticks_half_us = MXC_SYS_RiscVClockRate() / 1000000 / 3;
```
ttmut, line 45: Why `3`?

dnguye14-adi: It need a delay here to prevent CPU bother PCCR too much, then causes PCCR hardware cannot maintain the correct timing. After some testing I found this divisor (/3), it will give a delay less than 1us and big enough to prevent the PCCR. Here are my signal after add this delay, for both -O0, -O2 and -Og:

Picture1_1

Picture1_2

Picture1_3

sihyung-maxim, line 46: Adding a comment indicating why would be helpful for references.

ttmut (approved):

Libraries/PeriphDrivers/Source/SYS/mxc_delay.c
ttmut: Hi @sihyung-maxim, can you revisit?

lorne-maxim (approved):

PR #1380, opened 2025-04-21
fix(PeriphDrivers): Correct slot configuration loop in ADC SAR12

perkinsmg:
### Description

The loop in MXC_ADC_SlotConfiguration iterated one more than the number of slots passed in.
I can't think of any obvious reason why this is not incorrect.


sihyung-maxim (approved):
Nice catch! The ADC examples did not pass in the correct number of slots for these functions: `num_slots - 1` rather than `num_slots`. Instead of correcting the examples, these functions were "fixed".

Also found some minor errors with MAX32690 adc.h, so I'll create another PR that accounts for the effects of this PR.

Issue #699, opened 2023-08-06
FreeRTOS BLE "CordioM" task not receiving notifications


khpeterson:
While running repeats of my BLE streaming application (based on BLE_FreeRTOS), occasionally I end up in a state where the CordioM task is no longer able to receive notifications.  I traced this to inconsistent state in the FreeRTOS task control block and what seems like a race condition/synchronization bug.  I don't know how to fix the problem but I do know for sure that the CordioM task state variable ucNotifyState is not taskWAITING_NOTIFICATION even though the task is on the xSuspendedTaskList when notification is attempted.  

I instrumented xTaskGenericNotify to check for this condition:
```
            traceTASK_NOTIFY( uxIndexToNotify );

            /* If the task is in the blocked state specifically to wait for a
             * notification then unblock it now. */
            /********************************************************************/
	    /* tasks on delayed list must be waiting for notification */
	    if (listLIST_ITEM_CONTAINER( &( pxTCB->xStateListItem ) ) == &xSuspendedTaskList) {
		//configASSERT(ucOriginalNotifyState == taskWAITING_NOTIFICATION);
		if (ucOriginalNotifyState != taskWAITING_NOTIFICATION) {
			notify_out_of_sync_original_state = ucOriginalNotifyState;
			notify_out_of_sync_count++;
		}
	    }
            /********************************************************************/          
            if( ucOriginalNotifyState == taskWAITING_NOTIFICATION )

```
and then detect and report the condition in my app's run loop:
```
		extern volatile uint32_t notify_out_of_sync_count;
		extern volatile uint8_t notify_out_of_sync_original_state;
		extern uint32_t last_notify_out_of_sync_count;
		if (last_notify_out_of_sync_count != notify_out_of_sync_count) {
			TRACE_ERR2("detected notify out of sync, notify_out_of_sync_count = %d, original_state = %d", notify_out_of_sync_count, notify_out_of_sync_original_state);
			last_notify_out_of_sync_count = notify_out_of_sync_count;
		}
```
While running repeats, sometimes after just a few minutes, sometimes after several hours, I see in my console log:
```
khp@Kevins-MacBook-Pro-M1 % tail -f logs/repeats_20230805_ble_hang.log | grep "out of sync"
2023-08-05T18:18:51.211971: runLoop: detected notify out of sync, notify_out_of_sync_count = 1, original_state = 0
[...]

PR #1399, opened 2025-06-08
fix(Cordio/Wsf,FreeRTOS): Fix Issue #699 which results in hung Cordio/Wsf/FreeRTOS app

khpeterson:
See Issue #699 for details.

Bottom line, Wsf critical sections are not enforced because the critical nesting count is not shared between the port of Wsf to FreeRTOS and native FreeRTOS.  The result is unexpected behavior, and in the case of Issue #699, an early/unexpected yield from within a Wsf critical section eventually leads to corrupted CordioM task state and a hung application.

### DIFF
--- a/Libraries/Cordio/wsf/sources/targets/freertos/wsf_cs.c
+++ b/Libraries/Cordio/wsf/sources/targets/freertos/wsf_cs.c
@@ -34,9 +34,6 @@
   Global Variables
 **************************************************************************************************/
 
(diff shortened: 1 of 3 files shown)


khpeterson:
Any chance someone could review this soon?

The bug is pretty glaring/fundamental for anyone using BLE with FreeRTOS and the fix is small.

khpeterson:
Thank you for the review and merge!

PR #1403, opened 2025-06-13
fix(PeriphDrivers): Flush using a per-instance base address.

petejohanson-adi:
When flushing after a flash operation, we should be flushing a different base address depending on which flash instance was involved in the operation.

### Description

Found this issue while working on support for flash operations from the RV32 core on MAX32690, using the FLC1 instance, since the RV32 core can't really use FLC0. Without the change, the flush attempts to access the flash base address that isn't even accessible to the RV32, which causes an exception.

Work around this by passing which flash instance was part of the operation to `MXC_FLC_ME18_Flash_Operation` so it can determine what base address to use for flushing the cache.

### DIFF
--- a/Libraries/PeriphDrivers/Source/FLC/flc_me18.c
+++ b/Libraries/PeriphDrivers/Source/FLC/flc_me18.c
@@ -39,7 +39,7 @@
 /* **** Functions **** */
 
 //******************************************************************************
-void MXC_FLC_ME18_Flash_Operation(void)
+void MXC_FLC_ME18_Flash_Operation(mxc_flc_regs_t *flc)
 {
     /*
     This function should be called after modifying the contents of flash memory.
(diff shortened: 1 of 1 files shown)


Libraries/PeriphDrivers/Source/FLC/flc_me18.c
```diff
@@ -39,7 +39,7 @@
 /* **** Functions **** */
 
 //******************************************************************************
-void MXC_FLC_ME18_Flash_Operation(void)
+void MXC_FLC_ME18_Flash_Operation(mxc_flc_regs_t *flc)
```
ttmut, line 42: Should this be converted to `static` while we are at it?

ttmut (approved):

Libraries/PeriphDrivers/Source/FLC/flc_me18.c
petejohanson-adi: Done.

PR #1404, opened 2025-06-25
feat(Other, PeriphDrivers): Add ADC support for Zephyr stream

dimitrije-lilic:
Added functions are all new in order to not break anything existing. 
They are modification of existing functionalities but with 
different flags in order to achieve Zephyr stream with MAX32690 ADC.

### DIFF
--- a/Libraries/PeriphDrivers/Include/MAX32690/adc.h
+++ b/Libraries/PeriphDrivers/Include/MAX32690/adc.h
@@ -337,6 +337,20 @@ int MXC_ADC_StartConversion(void);
 int MXC_ADC_StartConversionAsync(mxc_adc_complete_cb_t callback);
 
 /**
+ * @brief   Perform a stream conversion on a specific channel
+ * @note    The channel must be configured separately
+ *          The ADC interrupt must be enabled and MXC_ADC_Handler() called in the ISR
+ *          places data in the error parameter of the callback function
+ *          Has different flags but behaves same as MXC_ADC_StartConversionAsync.
+ *          Enables interrupts for FIFO level only and starts ADC in continuous mode.
+ *
+ * @param   callback the function to call when the conversion is complete
+ *
+ * @return  Success/Fail, see \ref MXC_Error_Codes for a list of return codes.
+ */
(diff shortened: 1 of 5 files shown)


Libraries/PeriphDrivers/Include/MAX32690/adc.h
```diff
@@ -336,6 +336,19 @@ int MXC_ADC_StartConversion(void);
  */
 int MXC_ADC_StartConversionAsync(mxc_adc_complete_cb_t callback);
 
+/**
+ * @brief   Perform a stream conversion on a specific channel
+ * @note    The channel must be configured separately
+ *          The ADC interrupt must be enabled and MXC_ADC_Handler() called in the ISR
+ *          places data in the error parameter of the callback function
+ *          Has different flags but behaves same as MXC_ADC_StartConversionAsync.
```
ttmut, line 344: Can you explain a bit more in the description what is different than `MXC_ADC_StartConversionAsync`?

Libraries/zephyr/MAX/Include/wrap_max32_adc.h
```diff
@@ -284,6 +284,14 @@ static inline int Wrap_MXC_ADC_StartConversionAsync(uint32_t *sample_channels,
     return MXC_ADC_StartConversionAsync(callback);
 }
 
+static inline int Wrap_MXC_ADC_StartConversionAsyncStream(uint32_t *sample_channels,
+                mxc_adc_complete_cb_t callback)
+{
+    MXC_ADC_FIFO_Threshold_Config(MAX_ADC_FIFO_LEN >> 1);
+    Wrap_MXC_ADC_ChannelSelect(sample_channels);
+    return MXC_ADC_StartConversionAsyncStream(callback);
```
ttmut, line 292:
```suggestion

    return MXC_ADC_StartConversionAsyncStream(callback);
```

Libraries/PeriphDrivers/Include/MAX32690/adc.h
dimitrije-lilic: I agree, will add better explanation and update clang issues.

PR #1407, opened 2025-07-07
fix(Other): Only define ADC wrapper on MAX32690

petejohanson-adi:
Only define the ADC async stream wrapper on MAX32690 where it is supported.

### Description

This is needed to address a build regression introduced in #1404 on targets other than MAX32690.


ttmut (approved):
Not sure if an `#else` block that returns `E_NOT_SUPPORTED` should be added here.

MaureenHelm (approved):
@dimitrije-lilic FYI

hfakkiz:
> Not sure if an `#else` block that returns `E_NOT_SUPPORTED` should be added here.

If this function is going to be added to the Zephyr ADC driver, there should also be a wrapper function that returns -1 for other part numbers. Therefore, we need to add an `#else` block here and define also same function for MAX32655, MAX32665, MAX32680 and MAX78000.

hfakkiz:
> Ok, tweaked this to define a fallback wrapper that returns `E_NOT_SUPPORTED` on other targets.

I meant that this fallback wrapper should be included in the `#if` block between lines 56 and 151. That way, the function will also be defined for other part numbers (e.g., MAX32655, MAX32665, MAX32680, MAX78000), avoiding potential build failures on those targets.

PR #1414, opened 2025-07-23
ci(workflow): Add new Zephyr HAL workflow for fresh start

petejohanson-adi:
Improvement to our Zephyr export workflow, that will ensure we get discrete commits for each MSDK commit, which a clean history from the get-go. This will generate a new Zephyr HAL branch that will become the new source of truth in upstream Zephyr.

### Description

The [Zephyr ADI HAL repository](https://github.com/zephyrproject-rtos/hal_adi) has gotten out of sync with analogdevicesinc/hal_adi, in non-recoverable ways without breaking existing commits/history in messy ways. You can see the initial discussion on this on https://github.com/zephyrproject-rtos/hal_adi/pull/14

This PR is the first step to work to address that, by adding a new script for generating a new "clean" git history branch based on the MSDK commit stream, that also allows it to easily be re-run and add missing changes as additional discrete commits, without combining multiple MSDK commits into one HAL ADI commit, etc. It's based on the initial work on this by  @hfakkiz with many modifications.

Once merged, we would manually run the workflow via "workflow dispatch" in GHA, and verify the resulting branch is 100% correct. Assuming it is, we'd then work to get that new branch pulled into https://github.com/zephyrproject-rtos/hal_adi  and enforce that *any* changes to that repo are done as fast-forward pulls from the new generated HAL ADI, so MSDK stays the "source of truth" for the HAL ADI long term.


.github/workflows/scripts/zephyr_hal_sync.sh
```diff
@@ -0,0 +1,211 @@
...
+		dest="${hal_adi}/${NAME##Libraries/zephyr/}"
+		echo "New dest for Zephyr file is ${dest}"
+		mkdir -p $(dirname "${dest}")
+		cp "${msdk}/${NAME}" "${dest}"
+
+	fi
+    done < ${root_dir}/current_changed_files.txt
+
+    
+#    rm -rf ${hal_adi}/MAX/
```
hfakkiz: Please remove these comment lines.

.github/workflows/zephyr-hal-deprecated.yml
```diff
@@ -0,0 +1,90 @@
...
+          # Name the path/directory of this repo for the runner's workspacee
+          path: hal_adi
+      
+      - name: Copy necessary files from msdk to hal_adi
+        run: |
+          # Call or run any tasks here
+          # The working directory for scripts will be the root of the runner workspace.
+          # Using the checkout commands above, the script will start out in the same 
+          # directory that contains the 'msdk' and 'hal_adi' folders.
+          bash ./msdk/.github/workflows/scripts/zephyr-hal.sh
```
hfakkiz:
```suggestion
          bash ./msdk/.github/workflows/scripts/zephyr-hal-deprecated.sh
```
Might be nice to rename the old script as deprecated too.

.github/workflows/zephyr-hal.yml
```diff
@@ -36,27 +27,20 @@ jobs:
...
           fetch-depth: 0
-          # Name the path/directory of this repo for the runner's workspacee
           path: msdk
 
-      # Checkout another repository
+      # Checkout hal_adi repository
       - name: Checkout zephry hal repository
         uses: actions/checkout@v4
         with:
-          # Specific Branch, tag, or SHA to checkout (uses default branch if empty)
```
hfakkiz, line 52: Looks like `ref: 'develop'` isn't needed anymore.

.github/workflows/scripts/zephyr_hal_sync.sh
petejohanson-adi: Ooof. Silly oversight. Thanks.

.github/workflows/zephyr-hal-deprecated.yml
petejohanson-adi: Done.

.github/workflows/zephyr-hal.yml
petejohanson-adi: If you see the script, we pull a few "base" files from that branch to start the initial version of the export branch, so we might as well check that out first.

hfakkiz (approved):

.github/workflows/scripts/zephyr_hal_sync.sh
```diff
@@ -0,0 +1,164 @@
...
+echo "Starting to create/update history of hal_adi..."
+
+check_is_adi_hal_lib () {
+  [ "x$1" = x ] && return 1
+  for pf in Libraries/CMSIS/Include              \
+            Libraries/CMSIS/Device              \
+            Libraries/PeriphDrivers/Source              \
+            Libraries/PeriphDrivers/Include              \
+            Libraries/MAXUSB/include/core              \
+            Libraries/MAXUSB/src/core; do
```
MaureenHelm, line 92: What will happen if this list needs to change in the future? Say, for example, we decide to include svd files?

petejohanson-adi: A good question: Right now, this wouldn't easily have a recourse for adding those files from historical MSDK commits (since they would have already been processed). You would need some workaround where all the necessary files to include get touched somehow in a new commit, so the updated processing would pull in those files.

I'm not sure how we would avoid that without either:

* Re-processing from the start, which now re-writes history
* Have a different style of processing more like our current one, that just copies over everything we care about, and we see what changes. The *downside* to that is the end commits don't have the same "quality" since they're no longer "logical" commits like this script creates, but "bulk data import" commits.

MaureenHelm: Would it make sense in that case to introduce a new commit in the HAL repo that doesn't have a corresponding commit in the MSDK repo?

petejohanson-adi: Yeah, we could definitely do that. We'd need to tweak this script a bit to handle that; in particular we'd need to tweak the logic used to search for the "MSDK-Commit" git trailer to find the most recent commit that has such a trailer and use that as our starting point for "catching up".  I think that's probably a worthwhile addition here, so let me work on that.

petejohanson-adi: Ok, implemented with a bit of POSIX shell fun:

```
	iterator=0
	while [ -z "${hal_adi_current_msdk_sha}" ]
	do
		hal_adi_current_msdk_sha=$(git show --format="%(trailers:key=MSDK-Commit,valueonly,separator=%x2C)" --no-patch "HEAD~${iterator}")
		iterator=$(( $iterator + 1 ))
	done
```

Tested locally with my generated `msdk-export` branch, where I added an example "catch up" commit with `git commit -m "Test" --allow-empty` and then re-ran the updated script, and it properly searched for the catchup point and inspected the commits after that:

```
source ➤ sh zephyr_hal_sync.sh                                                                                                                                         
~/source/hal_adi ~/source
Already on 'msdk-sync'

Adding changes since 29c099f1141b694ecc5dad8d87fcb6f5cf5cd3b6 to the existing branch

~/source

Getting all commits of MSDK repository...

~/source/msdk ~/source
From https://github.com/analogdevicesinc/msdk
 * [new branch]            fix/ctb_reset -> origin/fix/ctb_reset
Trimming list to start after 29c099f1141b694ecc5dad8d87fcb6f5cf5cd3b6
~/source

Starting to create/update history of hal_adi...

Processing commit 1 of 2
~/source/msdk ~/source
Note: switching to '817feea27611594f089fc5d93e527c7b646f471b'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

HEAD is now at 817feea276 fix(BLE): Consolidate wsf/freertos and FreeRTOS native critical nesting count (fix issue #699, hanging Cordio/FreeRTOS app) (#1399)
~/source
~/source/hal_adi ~/source
No need to commits to hal_adi
~/source

Processing commit 2 of 2
~/source/msdk ~/source
Previous HEAD position was 817feea276 fix(BLE): Consolidate wsf/freertos and FreeRTOS native critical nesting count (fix issue #699, hanging Cordio/FreeRTOS app) (#1399)
HEAD is now at 7ddedf0e80 feat(Examples): Max78000 RF Segment Demo (#1387)
~/source
~/source/hal_adi ~/source
No need to commits to hal_adi
~/source
```

Issue #1289, opened 2024-12-09
MAX32655 - I2C driver hangs forever


lsampaio-adi:
Hi,
I am working with a MAX32655 custom board with 2 I2C devices on the same bus (I2C1). The devices are [ADXL343](https://www.analog.com/en/products/adxl343.html) and [MAX31328](https://www.analog.com/en/products/max31328.html). After experimenting with the I2C example, I noticed that sometimes the driver would just hang after running for a while.

When trying using the sync function `MXC_I2C_MasterTransaction`, it would not return. When trying to use the async function `MXC_I2C_MasterTransactionAsync` or `MXC_I2C_MasterTransactionDMA`, the interrupt would never come.

Using a logic analyzer, it seems to be a state machine issue with the I2C driver, as sometimes the STOP condition would never happen or the MAX32655 would not send an ACK.

As my board is based on the MAX32655FTHR, I tried to reproduce the issue with the 2 I2C components on the board (the Audio Codec and the PMIC). The main code below, when compiled for the Feather board, hangs. However, if you uncomment lines 124 and 149 (thus printing to the console), then it works.

I could not find a similar issue here with the I2C driver. Have you seen this problem before?

Thank you!

```c
#include <stdio.h>
#include <stdint.h>
#include <string.h>
#include "mxc_device.h"
#include "mxc_delay.h"
#include "nvic_table.h"
#include "i2c.h"
#include "dma.h"
#include "rtc.h"
#include "wut.h"

#define I2C_FREQ 400000
#define I2C_TCXO_ADDR 0x68
#define I2C_ACCEL_ADDR 0x53
#define I2C_BYTES 8

#define I2C_PMIC_ADDR 0b0101000
#define I2C_CODEC_ADDR 0b0011000

static uint8_t txdata[I2C_BYTES];
static uint8_t rxdata[I2C_BYTES];
volatile int I2C_FLAG;

void I2C_Callback(mxc_i2c_req_t *req, int error)
{
    I2C_FLAG = error;
    if (error) {
        printf("Callback: Error code: %d\n", error);
        while(1);
    }
}

void I2C1_IRQHandler(void)
{
    MXC_I2C_AsyncHandler(MXC_I2C1);
}

int main(void)
{
    mxc_i2c_req_t reqMaster;
    printf("\n******** I2C MAX31328 Test ********\n");

    MXC_Delay(MXC_DELAY_SEC(1));

    NVIC_EnableIRQ(I2C1_IRQn);

[...]

PR #1416, opened 2025-07-24
fix(PeriphDrivers): MAX32655 I2C driver hang forever (#1289)

dnguye14-adi:
### Description

Add a waiting for the last Transaction to finish and the Bus is ready.

### DIFF
--- a/Libraries/PeriphDrivers/Source/I2C/i2c_reva.c
+++ b/Libraries/PeriphDrivers/Source/I2C/i2c_reva.c
@@ -1035,6 +1035,9 @@ int MXC_I2C_RevA_MasterTransactionAsync(mxc_i2c_reva_req_t *req)
         return E_BAD_STATE;
     }
 
+    while (i2c->status & MXC_F_I2C_REVA_STATUS_BUSY) {}
+    // Wait for the last Transaction to finish and the Bus is ready
+
     if (AsyncRequests[i2cNum] == NULL) {
         if (req->addr > MXC_I2C_REVA_MAX_ADDR_WIDTH) {
             return E_NOT_SUPPORTED;


Libraries/PeriphDrivers/Source/I2C/i2c_reva.c
```diff
@@ -1035,6 +1035,9 @@ int MXC_I2C_RevA_MasterTransactionAsync(mxc_i2c_reva_req_t *req)
         return E_BAD_STATE;
     }
 
+    while (i2c->status & MXC_F_I2C_REVA_STATUS_BUSY) {}
+    // Wait for the last Transaction to finish and the Bus is ready
```
ttmut: This function is asynchronous, so it should ideally return immediately if a transfer cannot be initiated. However, there's a contradiction in behavior: at the beginning, the function blocks until the previous transaction finishes, while at the end, it returns `E_BUSY` if a transaction is already ongoing.

To maintain consistent behavior, consider moving the blocking logic under the conditional that checks if there is an active transaction (`AsyncRequests[i2cNum]`). This way, the function will wait for the busy flag to be cleared only if necessary and still exit early when a transfer can't start.

```c
if (AsyncRequests[i2cNum] == NULL) {
        if (req->addr > MXC_I2C_REVA_MAX_ADDR_WIDTH) {
            return E_NOT_SUPPORTED;
        }

        while (i2c->status & MXC_F_I2C_REVA_STATUS_BUSY) {}
        // Wait for the last Transaction to finish and the Bus is ready
```

dnguye14-adi: @ttmut ,
I updated as your suggestion, also verified the case,
Please help to review again,

Rgds,

ttmut: Thanks @dnguye14-adi

PR #1427, opened 2025-08-21
feat(PeriphDrivers, Other): Add clock calibration for MAX32662, Zephyr wrapper

trupples:
### Description

1. Implement `MXC_SYS_ClockCalibrate` for MAX32662, which performs the **IPO Calibration** procedure described in the [MAX32662 UG rev 1](https://www.analog.com/media/en/technical-documentation/user-guides/max32662-ug-7640.pdf) section 4.2.1.1. This adapts the already existing MAX32690 implementation added in https://github.com/analogdevicesinc/msdk/commit/3f96590c5f1a8b9f71d42e8f8569ebe4b08603a4
2. Add Zephyr wrapper that calls the MAX32662 or MAX32690 implementation, or returns `E_NOT_SUPPORTED` on other targets

Following this PR I intend to work on a PR in Zephyr to further expose this functionality.

#### Real use-case

I have found clock calibration necessary when using CAN on a MAX32662, where large enough differences between the clock frequencies of connected devices cause TX errors and ultimately devices dropping into BUS_OFF state. The datasheet doesn't specify the tolerance for IPO frequency, but I have empirically found it to be at least worse than 1%, which is already on the edge of what can be compensated for with CAN SJW. Running the auto-calibration routine before starting CAN communication seems to fix this issue.

#### Testing

I have tested this on three MAX32662 boards and can confirm that it builds and runs as expected.
To test the behaviour, I wrote a basic program that outputs a 5kHz PWM signal and measured the actual frequency of the pin.
Measurements before calibration: 5.005, 4.92, 4.97 kHz.
Measurements after calibration: 4.99, 5.01, 5.01 kHz.

This is a new API and thus introduces no backwards-incompatible changes.

### DIFF
--- a/Libraries/zephyr/MAX/Include/wrap_max32_sys.h
+++ b/Libraries/zephyr/MAX/Include/wrap_max32_sys.h
@@ -151,6 +151,15 @@ static inline int Wrap_MXC_SYS_GetUSN(uint8_t *usn)
 
 #endif // part number
 
+static inline int Wrap_MXC_SYS_ClockCalibrate(mxc_sys_system_clock_t clock)
+{
+#if defined(CONFIG_SOC_MAX32662) || defined(CONFIG_SOC_MAX32690)
+    return MXC_SYS_ClockCalibrate(clock);
+#else
+    return E_NOT_SUPPORTED;
+#endif
+}
+
 #ifdef __cplusplus
 }
 #endif


Libraries/zephyr/MAX/Include/wrap_max32_sys.h
```diff
@@ -149,6 +149,15 @@ static inline int Wrap_MXC_SYS_GetUSN(uint8_t *usn)
...
 }
 
+static inline int Wrap_MXC_SYS_ClockCalibrate(mxc_sys_system_clock_t clock)
+{
+#if defined(CONFIG_SOC_MAX32662) || defined(CONFIG_SOC_MAX32690)
+    return MXC_SYS_ClockCalibrate(clock);
+#else
+    return E_NOT_SUPPORTED;
+#endif
+}
```
ttmut, line 161: I think MAX32665, MAX32666 and MAX32650 are not covered by this function. The newly added wrapper falls under `#elif` part of the `#if` block that starts in line 37.
```c
#if defined(CONFIG_SOC_MAX32665) || defined(CONFIG_SOC_MAX32666) || defined(CONFIG_SOC_MAX32650)
...
    <--- // You need a Wrap_MXC_SYS_ClockCalibrate here as well
#elif defined(CONFIG_SOC_MAX32690) || defined(CONFIG_SOC_MAX32655) || \
    defined(CONFIG_SOC_MAX32670) || defined(CONFIG_SOC_MAX32672) ||   \
    defined(CONFIG_SOC_MAX32662) || defined(CONFIG_SOC_MAX32675) ||   \
    defined(CONFIG_SOC_MAX32680) || defined(CONFIG_SOC_MAX32657) ||   \
    defined(CONFIG_SOC_MAX78002) || defined(CONFIG_SOC_MAX78000) || defined(CONFIG_SOC_MAX32660)
...
static inline int Wrap_MXC_SYS_ClockCalibrate(mxc_sys_system_clock_t clock)
{
...
```

trupples: Oops, I didn't notice the big `#if`-`#elif` block didn't include all SoCs! I have moved the new function outside of it. I suppose having it all in one place is cleaner (?)

ttmut: That could work too, yeah.

PR #1429, opened 2025-08-26
fix(PeriphDrivers): ME18-191: Add code to workaround a power fail issue when entering and exiting Standby or UPM modes on the MAX32690

lorne-maxim:
### Description

Revision A4 and later of the device may inadvertently reset when exiting Standby or UPM if the power fail monitoring is enabled during these events.  The fix is to always disable the power fail reset just before entering one of these low power modes.  Once the part wakes up, the code then should wait for the power to be stable again before re-enabling the power fail monitor.

### DIFF
--- a/Libraries/PeriphDrivers/Source/LP/lp_me18.c
+++ b/Libraries/PeriphDrivers/Source/LP/lp_me18.c
@@ -28,15 +28,15 @@
 
 static void MXC_LP_DisablePowerFailDetection(void)
 {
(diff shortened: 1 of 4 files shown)


Libraries/PeriphDrivers/Source/LP/lp_me18.c
sihyung-maxim, line 1: Line 5, copyright update: 
` * Copyright (C) 2023-2025 Analog Devices, Inc.`

sihyung-maxim (approved):
Only thing is the copyright needs to be updated.

PR #1449, opened 2025-10-21
fix(PeriphDrivers): Fix UART static request upon initialization

rbudai98:
Initialize AsyncTxRequests and AsyncRxRequests arrays to NULL in MXC_UART_RevB_Init to prevent errors upon multiple initialization of same peripheral.

### Description

When the UART is reinitialized using asynchronous settings, a static variable retains previous requests, leading to a potential deadlock.

In particular, when initializing tinyIIOD from the no-OS repository, the system becomes stuck. This occurs because tinyIIOD configures the UART with specific settings without reading back characters. Upon attempting to reinitialize the same UART instance, the leftover requests from the previous initialization prevent the process from completing.

This fix ensures proper cleanup of pending UART requests during reinitialization to avoid deadlocks.

### DIFF
--- a/Libraries/PeriphDrivers/Source/UART/uart_revb.c
+++ b/Libraries/PeriphDrivers/Source/UART/uart_revb.c
@@ -167,6 +167,9 @@ int MXC_UART_RevB_Init(mxc_uart_revb_regs_t *uart, unsigned int baud, mxc_uart_r
     states[i].auto_dma_handlers = false;
     states[i].dma = NULL;
 
+    AsyncTxRequests[i] = NULL;
+    AsyncRxRequests[i] = NULL;
+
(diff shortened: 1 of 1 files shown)


rbudai98:
> Most UART settings can be configured using the existing API. What specific use case requires you to call `MXC_UART_Init` more than once?
> 
> If you're planning to reinitialize the UART, you can call `MXC_UART_AbortAsync` to cancel the ongoing transactions and reset request arrays. While the driver technically allows repeated calls to `MXC_UART_Init`, doing so without properly cleaning up the previous session may lead to unexpected behavior later on.

I took another look, and you’re right — the issue was on our side. We weren’t properly cleaning up before reinitialization.
I’ll go ahead and close this PR then.

PR #1456, opened 2025-10-29
fix(PeriphDrivers): MSDK-1337: CRC calculation issues

dnguye14-adi:
### Description

Fix CRC calculation issues and Update CRC examples.

_______________________________________________________________________________

### DIFF
--- a/Examples/MAX32670/CRC/main.c
+++ b/Examples/MAX32670/CRC/main.c
@@ -525,15 +525,15 @@ int run_demo_crc(int asynchronous)
             printf("PASSED!\n");
         } else {
             printf("FAILED!\n");
-			fail++;
+            fail++;
         }
 
         MXC_CRC_Shutdown();
 
         printf("\n");
     }
-	
-	return fail;
+
+    return fail;
 }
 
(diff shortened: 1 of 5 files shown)


Examples/MAX32670/CRC/main.c
```diff
@@ -33,97 +33,520 @@
...
-    if (fail != 0) {
-        printf("\nExample Failed\n");
-        return E_FAIL;
-    }
+    run_demo_crc(0);
+    run_demo_crc(1);
+
+    while (1) {}
 
-    printf("\nExample Succeeded\n");
```
ttmut, line 127: Instead of using an infinite loop, please implement a wait flag as in the previous version of this example so the code prints the test result once it completes. CI system relies on "Example Succeeded" string to determine the test result. See https://github.com/analogdevicesinc/msdk/blob/main/.github/workflows/scripts/example_tester_78002.sh#L222

ttmut (changes requested):

ttmut:
> CRC header and example for MAX32680 need to be updated as well.

This still needs to be addressed.

PR #1462, opened 2025-11-13
fix(PeriphDrivers): Fix LPWDT clock sources for MAX78002

perkinsmg:
From Aleks Kupriianov:
1. Before proposed changes, the implementation of WDT clock for 78002 put literally the enum value for clock to the corresponding register. This would be correct if the table in the page 384 of the user guide were correct.
2. However, it turns out that the clock diagram in the pdf is correct, and the sets and numbers of clocks are the same for 78000 and 78002. Henceforth, it makes sense to re-use the solution for 78000 for our case.
3. So we add 78002 to wdt_me17.c and use wdt_me17.c for 78002 as well (changing makefile). We already use it for 78000.
4. The files wdt_me17.c and wdt_ai87.c are nearly identical, except two places. One place is the function MXC_WDT_SetClockSource, where the substitution of enum is done. The other place is the function MXC_WDT_Init, which checks if the clock was enabled. It should be disabled in all chips, since LPGCR_PCLKDIS has ones for all chips.

### Description

Please include a summary of the changes and related issues. Please also include relevant motivation and context.


Libraries/PeriphDrivers/Source/WDT/wdt_me17.c
```diff
@@ -129,7 +129,7 @@ int MXC_WDT_SetClockSource(mxc_wdt_regs_t *wdt, mxc_wdt_clock_t clock_source)
     uint8_t idx = 0;
     uint8_t instance = 0;
 
-#if TARGET_NUM == 32655 || TARGET_NUM == 78000
+#if TARGET_NUM == 32655 || TARGET_NUM == 78000 || TARGET_NUM == 78002
     mxc_wdt_clock_t clock_sources[2][8] = {
         { MXC_WDT_PCLK, MXC_WDT_IBRO_CLK, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF },
         { MXC_WDT_IBRO_CLK, MXC_WDT_INRO_CLK, MXC_WDT_ERTCO_CLK, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF }
```
ttmut, line 135: How did you conclude that these mappings are correct for MAX78002?

ttmut:
> 4. The files wdt_me17.c and wdt_ai87.c are nearly identical, except two places. One place is the function MXC_WDT_SetClockSource, where the substitution of enum is done. The other place is the function MXC_WDT_Init, which checks if the clock was enabled. It should be disabled in all chips, since LPGCR_PCLKDIS has ones for all chips.
> 
According to the user manual, the reset value of `LPGCR_PCLKDIS` is 0x0 for MAX78002, and possibly for other chips as well. So unless this information is inaccurate, the `WDT1` clock is actually enabled on reset. I do not think re-enabling it will have any effect though.

PR #1463, opened 2025-11-13
fix(PeriphDrivers): Use separate buffer for USN AES calculation.

kevin-gillespie:
### Description

Use a separate buffer for the AES checksum verification. USN is only 13 bytes, but AES plain text buffer needs 16 bytes.

### DIFF
--- a/Libraries/PeriphDrivers/Source/SYS/sys_me30.c
+++ b/Libraries/PeriphDrivers/Source/SYS/sys_me30.c
@@ -43,6 +43,7 @@
 
 /* **** Definitions **** */
 #define MXC_SYS_CLOCK_TIMEOUT MSEC(1)
+#define MXC_SYS_USN_AES_LEN 16
 
 // DAP Lock macros
 #define INFOBLOCK_DAP_LOCK_OFFSET 0x30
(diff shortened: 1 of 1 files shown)


Libraries/PeriphDrivers/Source/SYS/sys_me30.c
```diff
@@ -85,12 +86,18 @@ int MXC_SYS_GetUSN(uint8_t *usn, uint8_t *checksum)
 
     /* If requested, verify and return the checksum */
     if (checksum != NULL) {
-        uint8_t check_csum[MXC_SYS_USN_CHECKSUM_LEN];
-        uint8_t aes_key[MXC_SYS_USN_CHECKSUM_LEN] = { 0 }; // NULL Key (per checksum spec)
+        uint8_t check_csum[MXC_SYS_USN_AES_LEN];
+        uint8_t check_usn[MXC_SYS_USN_AES_LEN] = { 0 }; // NULL plain text, add USN
```
petejohanson-adi, line 90: So... My understanding is the expectation of `MXC_SYS_GetUSN` is that callers, if requesting a checksum by passing a non-NULL `checksum` pointer, are expected to pass a larger allocated `usn` in, that is of size `MXC_SYS_USN_CHECKSUM_LEN` (which is defined as 16 everywhere, AFAICT), and this would avoid the need for an extra memcpy and putting an extra array on the stack.

Is there some reason we need to change that expectation?

kevin-gillespie: Ok I see, makes sense!.

PR #1465, opened 2025-11-20
fix(PeriphDrivers): Disable all unused CTB blocks prior to CTB operations and remove extraneous calls to MXC_CTB_InternalReset

lorne-maxim:
### Description

Certain cryptographic toolbox (CTB) operations may leave configuration bits set that can adversely affect future CTB operations. The PR makes sure those configuration bits are written to their default states prior to each and every operation type. This was achieved by adding a new API function (MXC_CTB_DisableBlocks) that clears the necessary bits. The enable bits for the one operation being performed are then configured appropriately.

This PR also removes most of the calls to MXC_CTB_InternalReset. This function destroys certain configuration parameters (like the cipher keys, for example) that are not stored anywhere else. Since the API cannot know when the application will set those particular configurations, it is left up to the application to manually call the MXC_CTB_InternalReset function instead of the API handling that automatically.

### DIFF
--- a/Libraries/PeriphDrivers/Source/CTB/ctb_reva.c
+++ b/Libraries/PeriphDrivers/Source/CTB/ctb_reva.c
@@ -1112,7 +1112,7 @@ static int MXC_CTB_Cipher_Generic(mxc_ctb_cipher_req_t *req, int op)
         // Wait until ready for data
         while (!MXC_CTB_Ready()) {}
 
-    	MXC_CTB_DoneClear(MXC_CTB_REVA_FEATURE_CIPHER);
+        MXC_CTB_DoneClear(MXC_CTB_REVA_FEATURE_CIPHER);
         MXC_CTB_DMA_StartTransfer(dataLength);
 
         // Wait until operation is complete


Libraries/PeriphDrivers/Source/CTB/ctb_reva.c
```diff
@@ -110,6 +113,14 @@ int MXC_CTB_RevA_Init(mxc_ctb_reva_regs_t *ctb_regs, uint32_t features)
     return E_NO_ERROR;
 }
 
+void MXC_CTB_RevA_DisableBlocks(mxc_ctb_reva_regs_t *ctb_regs)
+{
+    // Disable each of the individual CTB blocks. Only 1 block should be enable at a time.
+    ctb_regs->cipher_ctrl &= ~(MXC_F_CTB_REVA_CIPHER_CTRL_CIPHER);
+    ctb_regs->hash_ctrl &= ~(MXC_F_CTB_REVA_HASH_CTRL_HASH);
+    ctb_regs->crc_ctrl &= ~(MXC_F_CTB_REVA_CRC_CTRL_CRC);
```
ttmut, line 121:
```suggestion
    ctb_regs->crc_ctrl &= ~(MXC_F_CTB_REVA_CRC_CTRL_CRC);
    ctb_regs->crc_ctrl &= ~(MXC_F_CTB_REVA_CRC_CTRL_HAM);
```
Do we also need to disable ECC here as documented in the header?

```c
/**
 * @brief   This function is used by other API calls and should not be called 
 *          directly by application code.
 *
 * Disables the CRC, ECC, Hash, and Cipher blocks. Only one of these should
 * be enable at a time. 
 *
 */
void MXC_CTB_DisableBlocks(void);
```

lorne-maxim: Yes it should.  The PR has been updated accordingly.

PR #1486, opened 2025-12-25
fix(Examples): Update MAX78002 LPCMP example

TSenguler:
Updated the README file to reference the SW4 button instead of SW2.
Updated the printf statements in main.c to display SW4 instead of SW2.


Examples/MAX78002/LPCMP/main.c
ttmut, line 1: You should also update the copyright dates in this file.

TSenguler: Updated

PR #1545, opened 2026-04-20
fix(MAXUSB): Allow packet mode ReadEndpoint on EP0

petejohanson-adi:
### Description

Zephyr's USB stack issues a full size (MPS) OUT request on EP0 to help the stack avoid issues with not enough buffer allocated when hosts might accidentally send too much data. Adjust the MAXUSB library to allow packet mode to complete the request even if the MPS number of bytes hasn't be received (unlikely).

### DIFF
--- a/Libraries/MAXUSB/src/core/musbhsfc/usb.c
+++ b/Libraries/MAXUSB/src/core/musbhsfc/usb.c
@@ -692,7 +692,7 @@ static void event_out_data(uint32_t irqs)
         req->actlen += reqsize;
 
         if (!ep) {
-            if (req->actlen == req->reqlen) {
+            if ((req->actlen == req->reqlen) || reqsize < ep_size[0]) {
                 /* No more data */
                 MXC_USBHS->csr0 |= MXC_F_USBHS_CSR0_SERV_OUTPKTRDY | MXC_F_USBHS_CSR0_DATA_END;
                 /* Done */
@@ -1179,7 +1179,7 @@ int MXC_USB_ReadEndpoint(MXC_USB_Req_t *req)
 
             /* Signal to H/W that FIFO has been read */
             MXC_USBHS->csr0 |= MXC_F_USBHS_CSR0_SERV_OUTPKTRDY;
-            if (req->actlen == req->reqlen) {
+            if ((req->actlen == req->reqlen) || reqsize < ep_size[0]) {
                 /* Signal end of transaction to hardware */
                 MXC_USBHS->csr0 |= MXC_F_USBHS_CSR0_DATA_END;
 


Libraries/MAXUSB/src/core/musbhsfc/usb.c
```diff
@@ -621,7 +621,7 @@ static void event_out_data(uint32_t irqs)
     /* Loop for each data endpoint */
     for (ep = 0; ep < MXC_USBHS_NUM_EP; ep++) {
         buffer_bit = (1 << ep);
-        if ((irqs & buffer_bit) == 0) {
+        if ((irqs & buffer_bit) == 0 && ep != 0) {
```
hfakkiz: Why do we need to check EP0, even if the interrupt isn't related to it?

Libraries/MAXUSB/src/core/musbhsfc/usb.c
```diff
@@ -692,9 +692,13 @@ static void event_out_data(uint32_t irqs)
...
 
         if (!ep) {
-            if (req->actlen == req->reqlen) {
+            if ((req->type == MAXUSB_TYPE_PKT) || (req->actlen == req->reqlen)) {
                 /* No more data */
-                MXC_USBHS->csr0 |= MXC_F_USBHS_CSR0_SERV_OUTPKTRDY | MXC_F_USBHS_CSR0_DATA_END;
+                if (req->actlen == req->reqlen) {
+                    MXC_USBHS->csr0 |= MXC_F_USBHS_CSR0_SERV_OUTPKTRDY | MXC_F_USBHS_CSR0_DATA_END;
+                } else {
+                    MXC_USBHS->csr0 |= MXC_F_USBHS_CSR0_SERV_OUTPKTRDY;
```
hfakkiz: I noticed that data_end is not being set in csr0 here. Since we are still expecting more data, could you clarify why we are nulling the request prematurely without completing the transfer?

Libraries/MAXUSB/src/core/musbhsfc/usb.c
petejohanson-adi: We can probably remove this part, it's not 100% relevant to the fix needed in Zephyr: Fundamentally, this was *never* running for EP0, at least as I read it, since the flag never will get set for EP0. I'll remove though, so we can focus on the `ReadEndpoint` changes.

Libraries/MAXUSB/src/core/musbhsfc/usb.c
petejohanson-adi: Yeah, this is honestly the most contentious part, I've just pushed a tweak that reverts this part, which at least seems to work on two different hosts here fine.

I think the *only* risk, which may not actually ever happen in reality, would be a host sending only a *partial* amount of the full data in the first packet. I don't believe this is even possible, but could be a risk?

hfakkiz: Thanks for the revert. While you're concerned about partial packets, the more immediate risk is the multi-packet transfer scenario. Since transfers exceeding the MPS are split, triggering data_end after the first packet causes a state mismatch. The host still has remaining data to send as defined by wLength, and the device closing the transfer prematurely leads to a STALL or timeout. We should ensure the state machine tracks completion against wLength to guarantee the full transfer is received.

hfakkiz (approved):
Yes, we can terminate the transfer if the received packet size is smaller than the EP0 MPS, as this signals the end of data.

PR #1566, opened 2026-05-28
fix(PeriphDrivers): Fix clock speed configuration of SPI3 for MAX32650

cameronpacileo-adi:
### Description

spi_me10.c assumes the SPI3 peripheral clock is twice the frequency as other SPI instances. Looking at the [MAX32650 UG](https://www.analog.com/media/en/technical-documentation/user-guides/max32650max32652-user-guide.pdf) there is no indication SPI3 peripheral clock is twice the frequency.

### DIFF
--- a/Libraries/PeriphDrivers/Source/SPI/spi_me10.c
+++ b/Libraries/PeriphDrivers/Source/SPI/spi_me10.c
@@ -123,9 +123,7 @@ int MXC_SPI_ReadyForSleep(mxc_spi_regs_t *spi)
 /* ************************************************************************ */
 int MXC_SPI_GetPeripheralClock(mxc_spi_regs_t *spi)
 {
-    if (MXC_SPI_GET_IDX(spi) == 3) {
-        return PeripheralClock * 2;
-    } else if (MXC_SPI_GET_IDX(spi) != -1) {
+    if (MXC_SPI_GET_IDX(spi) != -1) {
         return PeripheralClock;
     } else {
         return E_BAD_PARAM;


hfakkiz (approved):

lorne-maxim:
Were the changes of PR #1566 tested on actual hardware?  On the MAX32650, SPI3 is on the AHB bus not the APB bus and is therefore clocked twice as fast as the other SPI peripherals.  This can be confirmed in Table 2-3 of the user guide.  These changes may need to be reverted.

ttmut:
> Were the changes of PR #1566 tested on actual hardware? On the MAX32650, SPI3 is on the AHB bus not the APB bus and is therefore clocked twice as fast as the other SPI peripherals. This can be confirmed in Table 2-3 of the user guide. These changes may need to be reverted.

Makes sense but there is conflicting information in Table 19-7 (scaling formula uses PCLK). Also, in Table 1 in datasheet the maximum frequency is the same for all SPI instances.

lorne-maxim:
I think you (and the PR) may be correct).  I've checked with a logic analyzer and having clocked from PCLK seems to be correct.  I am still investigating, but you can ignore my comment for now.

ttmut:
Thanks for the clarification.

lorne-maxim:
After further investigation, it turns out the HW is internally scaling the values written to the clk_cfg register.  Although SPI3 is on the AHB, the resulting output SPI clock is half of what should be expected given the values written to that register.  Here are the exact changes the HW does internally for the fields in the clk_cfg register:

image
 
The bottom line -- This PR is correct as is.

PR #1568, opened 2026-06-02
fix(Build): Share multi core data 7800x when using RISCV_LOAD

perkinsmg:
### Description

The SystemCoreClock variable, placed into the .shared section, was only placed at identical addresses in the Cortex and RISC-V projects when using the _arm.ld and _rv ld scripts together. When using the max7800x.ld script and the RISCV_LOAD mechanism, the variable was placed at different addresses in both executables.

Fix this, by creating a SHARED memory region and placing the data there.

Note that there is a subtlety with how I do this: in CFS, we remove the MEMORY command from these .ld scripts and CFS provides it. Since old CFS projects won't have the SHARED memory region, we would get a link error if we map to that directly. So instead, define a symbol to that address, and map to that address instead. Use an assertion to check the data fits. Then, for old projects, we'll map the variable right to the end of memory if SHARED is not defined, otherwise CFS will provide the symbols with the address of the SHARED region defined there.

### DIFF
--- a/Libraries/CMSIS/Device/Maxim/MAX78000/Source/GCC/common.ld
+++ b/Libraries/CMSIS/Device/Maxim/MAX78000/Source/GCC/common.ld
@@ -1,5 +1,6 @@
 /* Shared RISC-V and ARM memory map */
 _FlashLength_ =  __FlashLength;
+
 MEMORY {
     ROM     (rx)  : ORIGIN = 0x00000000, LENGTH = 0x00010000 /* 64kB - ROM */
     
@@ -9,8 +10,14 @@ MEMORY {
 
     SRAM1   (rwx) : ORIGIN = 0x20008000, LENGTH = 0x00008000 /* 32 kB - ARM Stack */
     
-    SRAM2   (rwx) : ORIGIN = 0x20010000, LENGTH = 0x0000C000 /* 48kB SRAM2 - RV Data */
+    SRAM2   (rwx) : ORIGIN = 0x20010000, LENGTH = 0x0000BFF8 /* ~48kB SRAM2 - RV Data */
+
+    SHARED  (rw)  : ORIGIN = 0x2001BFF8, LENGTH = 0x00000008 /* 8B Shared - ARM, RV Data */
     
     SRAM3   (rwx) : ORIGIN = 0x2001C000, LENGTH = 0x00004000 /* 16kB SRAM3 - RV Code */
 }
+
+/* The address and length of the SHARED section defined above. */
+__shared_origin = ORIGIN(SHARED);
+__shared_length = LENGTH(SHARED);
 
--- a/Libraries/CMSIS/Device/Maxim/MAX78002/Source/GCC/max78002_arm.ld
+++ b/Libraries/CMSIS/Device/Maxim/MAX78002/Source/GCC/max78002_arm.ld
@@ -2,7 +2,7 @@
  *
  * Copyright (C) 2022-2023 Maxim Integrated Products, Inc. (now owned by 
  * Analog Devices, Inc.),
- * Copyright (C) 2023-2024 Analog Devices, Inc.
+ * Copyright (C) 2023-2026 Analog Devices, Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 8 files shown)


Libraries/CMSIS/Device/Maxim/MAX78002/Source/GCC/max78002_arm.ld
```diff
@@ -52,15 +52,15 @@ SECTIONS {
...
+    .shared (NOLOAD) :
     {
         . = ALIGN(4);
         _shared = .;
         *(.mailbox*)
         . = ALIGN(4);
         *(.shared*)     /*read-write zero initialized data: uninitialzed global variable*/
         _eshared = ALIGN(., 4);
-    } > SRAM5 AT>FLASH
+    } > SHARED
```
perkinsmg, line 63: Note that the effect here is that this has moved from SRAM5 to SRAM7. I don't see any reason why that wouldn't work but I'm suspicious.

petejohanson-adi: This would affect things if ICC1 is enabled, AFAIK.

Image

perkinsmg: Good catch, thanks. Returned it to SRAM5 (and similarly on MAX78000).

perkinsmg: Hmm, actually I didn't do that quite right. Marking draft and will revise tomorrow.

perkinsmg: Revised.

petejohanson-adi:
No objections, but might be nice to find someone who's more familiar with the bare metal / MSDK bits to validate as well, since my experience with the linking details is more limited, since our Zephyr integration doesn't use the MSDK linker scripts.

Libraries/CMSIS/Device/Maxim/MAX78000/Source/GCC/common.ld
```diff
@@ -9,7 +14,9 @@ MEMORY {
 
     SRAM1   (rwx) : ORIGIN = 0x20008000, LENGTH = 0x00008000 /* 32 kB - ARM Stack */
     
-    SRAM2   (rwx) : ORIGIN = 0x20010000, LENGTH = 0x0000C000 /* 48kB SRAM2 - RV Data */
+    SRAM2   (rwx) : ORIGIN = 0x20010000, LENGTH = 0x0000BFF8 /* ~48kB SRAM2 - RV Data */
+
+    SHARED  (rw)  : ORIGIN = 0x2001BFF8, LENGTH = 0x00000008 /* 8B Shared - ARM, RV Data */
```
ako-adi, line 15: Could use the symbols defined above here, to ensure consistency. (The values here can also be expressions.)

perkinsmg: Thanks for the prod. I don't want to do it that way round, as I think it would make it easy to accidentally define a region that overlaps with the other region definitions.
But I should be able to use ORIGIN and LENGTH to create my symbols from the region definition I think.

ako-adi: That is a very good idea.

perkinsmg: Fixed.

perkinsmg:
Hi @ttmut  - would you be able to review this? Also, any clues as to why the tests are failing? One is for MAX32690, which hasn't changed here. The other is something to do with git but as far as I could see the suggested change is present.

ttmut:
> Hi @ttmut - would you be able to review this? Also, any clues as to why the tests are failing? One is for MAX32690, which hasn't changed here. The other is something to do with git but as far as I could see the suggested change is present.

Build tests are failing because your branch is behind `analogdevicesinc/main`. Hardware tests failed due to an OpenOCD failire. I think we need a retry mechanism for such cases.

ttmut (approved):

ttmut:
Can you check the failures in https://github.com/analogdevicesinc/msdk/actions/runs/28154402216/job/83379363303?

perkinsmg:
> Can you check the failures in https://github.com/analogdevicesinc/msdk/actions/runs/28154402216/job/83379363303?

I can see those are for MAX78000, so I'll take a look.

PR #1603, opened 2026-08-17
feat(PeriphDrivers): Allow skipping RTC clock config for all parts

ttmut:
### Description

Continuation of https://github.com/analogdevicesinc/msdk/pull/1601

Wrap clock configuration code in `MXC_RTC_Init` with `MSDK_NO_GPIO_CLK_INIT` guard so that clock configuration can be done externally.

In Zephyr side, clock source selection for RTC will be done outside the RTC driver, so remove the obsolete wrapper code. Also add external clock source option to `wrap_get_32k_clock_source_instance`.

With https://github.com/zephyrproject-rtos/zephyr/pull/116541, we can support `RTC_CLK_IN` as 32kHz clock source.

### DIFF
--- a/Libraries/PeriphDrivers/Include/MAX32570/rtc.h
+++ b/Libraries/PeriphDrivers/Include/MAX32570/rtc.h
@@ -104,6 +104,8 @@ int MXC_RTC_Stop(void);
 
 /**
  * @brief     Initialize the sec and ssec registers and enable RTC (Blocking function)
+ * @note      If you wish to manage clock configuration externally, define the flag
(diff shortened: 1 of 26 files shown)


petejohanson-adi (approved):

ttmut:
> 
> In Zephyr side, clock source selection for RTC will be done outside the RTC driver, so remove the obsolete wrapper code. Also add external clock source option to `wrap_get_32k_clock_source_instance`.
> 

Restored `wrap_max32_rtc.h` for now as it could break upstream Zephyr support in case the PR gets rejected. We can remove it once Zephyr changes are accepted.

PR #1624, opened 2026-09-15
fix(PeriphDrivers): Initialize SPI DMA channel states before shutdown

haluk-k:
## Description

Initialize the SPI RevA1 TX and RX DMA channel state fields to
`E_NO_DEVICE` during static initialization.

The Zephyr MAX32 SPI driver calls `MXC_SPI_Shutdown()` before the first
SPI initialization. Since the static SPI state array was zero-initialized,
both DMA channel fields initially contained channel `0`.

As a result, `MXC_SPI_RevA1_Shutdown()` treated DMA channel 0 as an
acquired channel and released it. A subsequent SPI DMA transfer attempted
to use the released channel and returned `E_BAD_PARAM`.

## Changes

- Initialize `channelTx` to `E_NO_DEVICE` for every SPI instance
- Initialize `channelRx` to `E_NO_DEVICE` for every SPI instance
- Preserve the existing initialization and shutdown behavior

## Root Cause

The default zero initialization of the static SPI state array represents a
valid DMA channel number. However, before `MXC_SPI_RevA1_InitState()` is
called, no DMA channel has actually been acquired by the SPI driver.

Calling shutdown during this state therefore releases DMA channel 0
incorrectly.

## Expected Result

The initial SPI shutdown does not release DMA channel 0 unless the SPI
driver previously acquired it. SPI DMA transfers can subsequently use the
configured DMA channels without returning E_BAD_PARAM.


haluk-k:
Hi @ttmut , could you please take a look at this PR? I wasn't able to assign you as a reviewer. Thanks!

ttmut:
Hi @haluk-k, I believe this is not necessary anymore since https://github.com/zephyrproject-rtos/zephyr/commit/eeef2b9042946aef0d5941971f80225cc1e56fa0

haluk-k:
Hi @ttmut, I think the change you mentioned will solve the problem. Therefore, I will close this PR, since it seems to be unnecessary.

